PR #10, opened 2022-06-06
Update to work with latest `hydrogen-view-sdk@0.0.12`

MadLittleMods:
Update to work with latest `hydrogen-view-sdk@0.0.12` after there has been months of changes.

Also nice to develop with the SDK now that https://github.com/vector-im/hydrogen-web/pull/693 has merged.

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -23,6 +23,8 @@ module.exports = defineConfig({
     },
   },
   build: {
+    sourcemap: 'inline',
+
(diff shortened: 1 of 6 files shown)


vite.config.js
```diff
@@ -23,6 +23,8 @@ module.exports = defineConfig({
     },
   },
   build: {
+    sourcemap: 'inline',
```
MadLittleMods, line 26: This doesn't seem to do much but I would like the errors to look better.

Will revert here and try to introduce a new PR that actually has an effect.

PR #14, opened 2022-06-08
Organize `start-dev` file and add more `nodemon` logging

MadLittleMods:
Organize `start-dev` file and add more `nodemon` logging

Spawned from https://github.com/vitejs/vite/issues/8492

### DIFF
--- a/server/start-dev.js
+++ b/server/start-dev.js
@@ -1,5 +1,7 @@
 'use strict';
 
+console.log('start-dev process.env.NODE_ENV', process.env.NODE_ENV);
+
 const path = require('path');
 const nodemon = require('nodemon');
 const { build } = require('vite');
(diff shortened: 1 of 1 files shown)


server/start-dev.js
```diff
@@ -1,13 +1,23 @@
...
-console.log('start-dev process.env.NODE_ENV', process.env.NODE_ENV);
+// Build the client-side JavaScript bundle when we see any changes
+build(
+  mergeOptions(viteConfig, {
+    build: {
+      // Rebuild when we see changes
+      watch: true,
+    },
+  })
+);
```
MadLittleMods, line 20: Just putting the Vite build above so it's close to the `viteConfig` that we import

server/start-dev.js
```diff
@@ -30,16 +40,13 @@ nodemon
...
-      // Rebuild when we see changes
-      watch: true,
-    },
   })
-);
-
-console.log('start-dev2 process.env.NODE_ENV', process.env.NODE_ENV);
+  .on('crash', function () {
+    console.log('Nodemon: script crashed for some reason');
+  })
```
MadLittleMods, line 46: Found a new event to keep an eye on

server/start-dev.js
```diff
@@ -30,16 +40,13 @@ nodemon
...
+  // .on('watching', (file) => {
+  //   console.log('watching');
+  // })
+  .on('log', function (data) {
+    console.log(`Nodemon logs: ${data.type}: ${data.message}`);
+  });
```
MadLittleMods, line 52: This will give us all of the `--verbose` logging you would normally see using the CLI

PR #15, opened 2022-06-09
Remove `matrix-bot-sdk` usage in tests

MadLittleMods:
Remove `matrix-bot-sdk` usage in tests because it didn't have timestamp massaging `?ts` and it's not really necessary to rely on since we can just call the API directly 🤷. `matrix-bot-sdk` is also very annoying having to build rust crypto packages.

We're now using direct `fetch` requests against the Matrix API and lightweight `client` object.

```sh
npm run test

> matrix-public-archive@0.0.0 test
> npm run mocha -- test/e2e-tests.js --timeout 15000


> matrix-public-archive@0.0.0 mocha
> mocha "test/e2e-tests.js" "--timeout" "15000"

Config is using nodeEnv=test


  matrix-public-archive
server process.env.NODE_ENV test
    Test fixture homeservers
      √ Test federation between fixture homeservers (1864ms)
    Archive
      √ shows all events in a given day (1020ms)
      √ can render diverse messages (1193ms)
      - can render day back in time from room on remote homeserver we haven't backfilled from
      - will redirect to hour pagination when there are too many messages
      - will render a room with only a day of messages
      - will render a room with a sparse amount of messages (a few per day) with no contamination between days


  3 passing (4s)
  4 pending
```

Here is the interactive result at the end of the diverse messages test. Everything is working as expected and image is thumbnailing and full image in lightbox,


test/e2e-tests.js
```diff
@@ -218,84 +204,103 @@ describe('matrix-public-archive', () => {
...
+            h: 312,
+            'xyz.amorgan.blurhash': 'LkR3G|IU?w%NbxbIemae_NxuD$M{',
           },
-          w: 1894,
-          h: 925,
-          'xyz.amorgan.blurhash': 'LkR3G|IU?w%NbwbIemae_NxuD$M{',
-          // TODO: How to get a proper thumnail URL that will load?
-          thumbnail_url: mxcUri,
+          msgtype: 'm.image',
+          // TODO: How can we get the thumbnail to load? The full lightbox version works.
```
MadLittleMods, line 231: We need to enable `dynamic_thumbnails` in `homserver.yaml`. This was a bit non-obvious to know because Synapse was just throwing out 404's before. I've created an issue to make the error more obvious and friendly, https://github.com/matrix-org/synapse/issues/13016

PR #20, opened 2022-06-15
Add testing to CI

MadLittleMods:
Add testing to CI

 1. Build test homeserver Docker images which can federate with each other
 1. Run end-to-end (e2e) tests

## Dev notes

 - https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions
 - https://docs.github.com/en/actions/learn-github-actions/contexts
 
Sharing variables across jobs when the `services` field can't access the `env` context:

 - https://github.community/t/how-to-use-env-with-container-image/17252/24


package.json
```diff
@@ -29,7 +30,7 @@
   },
   "dependencies": {
     "express": "^4.17.2",
-    "hydrogen-view-sdk": "^0.0.4",
+    "hydrogen-view-sdk": "npm:@mlm/hydrogen-view-sdk@^0.0.12",
```
MadLittleMods, line 33: Custom version `hydrogen-view-sdk` which runs the scratch changes from https://github.com/vector-im/hydrogen-web/pull/653 so everything just works out of the box

PR #22, opened 2022-06-15
Containerize app (make a Docker image for `matrix-public-archive`)

MadLittleMods:
Containerize app (make a Docker image for `matrix-public-archive`)

```
ghcr.io/matrix-org/matrix-public-archive/matrix-public-archive
```

... with a variety of tags!

## Dev notes

 - https://github.com/orgs/matrix-org/packages?tab=packages&q=matrix-public-archive

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -56,14 +56,14 @@ jobs:
         uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
         with:
           images: ${{ steps.save_var.outputs.docker_image_name }}
+          # Defaults (as indicated by https://github.com/docker/metadata-action/tree/97c170d70b5f8bfc77e6b23e68381f217cb64ded#tags-input).
+          # Plus custom tags:
+          #  - Full length sha
           tags: |
-            # Defaults (as indicated by https://github.com/docker/metadata-action/tree/97c170d70b5f8bfc77e6b23e68381f217cb64ded#tags-input) 
             type=schedule
             type=ref,event=branch
             type=ref,event=tag
             type=ref,event=pr
-            # Custom tags:
-            #  - Full length sha
             type=sha,format=long
 
       - name: Build and push Docker image
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -86,7 +86,7 @@ jobs:
     services:
       # We need two homeservers that federate with each other to test with
       hs1:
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yaml
```diff
@@ -47,21 +47,36 @@ jobs:
...
+          images: ${{ steps.save_var.outputs.docker_image_name }}
+          # Defaults (as indicated by https://github.com/docker/metadata-action/tree/97c170d70b5f8bfc77e6b23e68381f217cb64ded#tags-input).
+          # Plus custom tags:
+          #  - Full length sha
+          tags: |
+            type=schedule
+            type=ref,event=branch
+            type=ref,event=tag
+            type=ref,event=pr
+            type=sha,format=long
```
MadLittleMods, line 67: Re-introducing the `docker/metadata-action` so that we get nice tags for PRs, branches, and tags. It just does the right thing.

We additionally added `type=sha,format=long` the default list of tags so that we can still reference it in `services` below. The only change is that it uses the `sha-xxx` prefix for the commit sha tags.

.github/workflows/tests.yaml
```diff
@@ -47,21 +47,36 @@ jobs:
       # the next job.
       #
       # via https://github.community/t/how-to-use-env-with-container-image/17252/25
-      - name: Save the Docker image name to a variable so we can share and re-use it in other jobs via `${{ needs.deploy-test-synapse-image.outputs.docker_image_name }}`
+      - name: Save the Docker image name to a variable so we can share and re-use it in other jobs via `${{ needs.build-test-synapse-image.outputs.docker_image_name }}`
```
MadLittleMods, line 50: Removing the tag from the end of the name.

We now specify this explicitly set this in the `services` which still works 🤷

test/dockerfiles/synapse/log_config.yaml
```diff
@@ -17,6 +17,11 @@ handlers:
     # log to stdout, for easier use with 'docker logs'
     stream: 'ext://sys.stdout'
 
+
+# loggers:
+#     synapse.rest.media.v1.thumbnail_resource:
+#         level: DEBUG
```
MadLittleMods, line 23: This isn't really related to this PR but was messing around with this on https://github.com/matrix-org/synapse/issues/13016 and https://github.com/matrix-org/synapse/issues/13039 and thought it was good reference whenever we want to add logging for something else down the line. Saves me from looking it up to copy-paste.

PR #23, opened 2022-06-15
Make sure container is able to start up

MadLittleMods:
Make sure container is able to start up.

Follow-up to https://github.com/matrix-org/matrix-public-archive/pull/22

## Dev notes

```
docker build -t matrix-public-archive .
docker run -d -p3053:3050 --env matrixServerUrl=http://FAKE_SERVER/ --env matrixAccessToken=FAKE_TOKEN matrix-public-archive
```

---

https://stackoverflow.com/questions/42737957/how-to-view-docker-compose-healthcheck-logs
```
docker inspect --format "{{json .State.Health }}" wizardly_mcclintock
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,10 @@ RUN mkdir -p /app
 
 WORKDIR /app
 
-RUN apt-get update && apt-get install -y curl
-
 RUN npm install npm@^8 --location=global
+
+# Copy the health-check script
+COPY docker-health-check.js /app/
 
 # Copy just what we need to install the dependencies so this layer can be cached
 # in the Docker build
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -21,4 +24,6 @@ RUN npm run build
 # Copy the rest of the app
 COPY server /app/server/
 
-ENTRYPOINT ["npm" "start"]
+HEALTHCHECK CMD curl --fail http://localhost:3050/health-check
```
MadLittleMods, line 27: Lots of the examples out there add a `|| exit 1` to the end. Does the Docker `HEALTHCHECK` really not play well with non-`1` error codes?

MadLittleMods: Maybe:

> - 0: success - the container is healthy and ready for use
> - 1: unhealthy - the container is not working correctly
> - 2: reserved - do not use this exit code
>
> *-- https://docs.docker.com/engine/reference/builder/#healthcheck*

We could move to a custom health check Node.js script and ditch the `curl`

MadLittleMods: Moved to a Node.js script and removed the `curl` dependency ⏩

PR #24, opened 2022-06-15
Fix image not running in K8s (ROSA)

MadLittleMods:
Fix image not running in K8s (ROSA)

Trying to solve the container exiting with error code `243`. `npm` throws `243` for any command you try to run with it.

Here is how to get into the container while the entrypoint normally errors.
```
$ oc run -i --tty --image=ghcr.io/matrix-org/matrix-public-archive/matrix-public-archive:sha-65edaea1a9713bb2cc93fa98638327fdeff765cd mpa-test2 --port=3050 --restart=Never --env="DOMAIN=cluster" --command /bin/bash
$ npm start

$ echo $?
243
```

Why does the same image work in Docker?
```
$ docker run -it --entrypoint /bin/bash ghcr.io/matrix-org/matrix-public-archive/matrix-public-archive:sha-65edaea1a9713bb2cc93fa98638327fdeff765cd
$ npm start
# it starts ✅
```

## Root cause

Seems like it could be either of these issues. Both are `8.6.0`+ problems as well which lines up with the `npm` version available in each base image.

 - https://github.com/npm/cli/issues/4769
 - https://github.com/npm/cli/issues/4996
    - Specific comment about breaking stuff in k8s for other people, https://github.com/npm/cli/issues/4996#issuecomment-1150290804

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:16.15-buster-slim
+FROM node:16.14.2-buster-slim
 
 RUN mkdir -p /app
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,7 @@
+# XXX: Before updating this, make sure the issues around `npm` silently exiting
+# with error 243 issues are solved:
+#  - https://github.com/npm/cli/issues/4996
+#  - https://github.com/npm/cli/issues/4769
 FROM node:16.14.2-buster-slim
 
 RUN mkdir -p /app
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,8 +4,6 @@ RUN mkdir -p /app
 
 WORKDIR /app
 
-RUN npm install npm@^8 --location=global
-
 # Copy the health-check script
 COPY docker-health-check.js /app/
 


Dockerfile
```diff
@@ -4,8 +4,6 @@ RUN mkdir -p /app
 
 WORKDIR /app
 
-RUN npm install npm@^8 --location=global
```
MadLittleMods, line 7: Doesn't solve the problem but I think we can get rid of this anyway. I added it before because I saw a warning about the `package-lock.json` made in an earlier version and to do the one-time upgrade, etc.

Dockerfile
```diff
@@ -1,11 +1,9 @@
-FROM node:16.15-buster-slim
+FROM node:16.14.2-buster-slim
```
MadLittleMods, line 1: This seems to have fixed it 🤷 . Here are the packaged versions:

```sh
$ node --version
v16.14.2
$ npm --version
8.5.0
```

Previously it was using:

```sh
$ node --version
v16.15.1
$ npm --version
8.11.0
```

MadLittleMods: Seems like it could be either of these issues. Both are `8.6.0`+ problems as well which lines up with the `npm` version available in each base image.

 - https://github.com/npm/cli/issues/4769
 - https://github.com/npm/cli/issues/4996

PR #27, opened 2022-06-17
Draft: OpenTelemetry tracing

MadLittleMods:


server/tracer.js
```diff
@@ -0,0 +1,87 @@
...
+    getNodeAutoInstrumentations({
+      '@opentelemetry/instrumentation-http': {
+        // Docs:
+        // https://github.com/open-telemetry/opentelemetry-js/tree/51afd54bd63e46d5d530266761144c7be2f6b3a7/experimental/packages/opentelemetry-instrumentation-http
+        //
+        // This is the place to ignore root level spans for Express routes. My
+        // first guess would be in `ignoreLayers` in
+        // `@opentelemetry/instrumentation-express` but that's not the case, see
+        // https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1034#issuecomment-1158435392
+        ignoreIncomingPaths: [
```
MadLittleMods, line 59: `ignoreIncomingPaths` is deprecated

We can switch to using `ignoreIncomingRequestHook` here, https://github.com/open-telemetry/opentelemetry-js/tree/51afd54bd63e46d5d530266761144c7be2f6b3a7/experimental/packages/opentelemetry-instrumentation-http#http-instrumentation-options

MadLittleMods: Done ✅

config/config.default.json
```diff
@@ -3,6 +3,8 @@
   "basePath": "http://localhost:3050",
   "matrixServerUrl": "http://localhost:8008/",
   "archiveMessageLimit": 500,
+  "requestTimeoutMs": 25000,
+  "UNCOMMENT_jaegerTracesEndpoint": "http://localhost:14268/api/traces",
```
MadLittleMods, line 7: Would be good to introduce `jsonc` to allow comments in the config.

I would like to display this config option without having it on by default since it requires a third-party service (Jaeger).

package.json
```diff
@@ -30,6 +30,17 @@
...
+    "@opentelemetry/auto-instrumentations-node": "^0.31.0",
+    "@opentelemetry/context-async-hooks": "^1.4.0",
+    "@opentelemetry/core": "^1.4.0",
+    "@opentelemetry/exporter-jaeger": "^1.3.1",
+    "@opentelemetry/instrumentation": "^0.30.0",
+    "@opentelemetry/propagator-ot-trace": "^0.26.0",
+    "@opentelemetry/resources": "^1.3.1",
+    "@opentelemetry/sdk-trace-base": "^1.3.1",
+    "@opentelemetry/semantic-conventions": "^1.3.1",
+    "dompurify": "^2.3.9",
```
MadLittleMods, line 43: Same HTML sanitizer that Hydrogen uses, [ttps://github.com/vector-im/hydrogen-web/blob/.../src/platform/web/parsehtml.js#L66](https://github.com/vector-im/hydrogen-web/blob/c9bca52e82d96ae3545b3623c1a31522ddb30d91/src/platform/web/parsehtml.js#L66)

Link to project: https://github.com/cure53/DOMPurify

server/lib/safe-json.js
```diff
@@ -0,0 +1,10 @@
+'use strict';
+
+// via https://gitlab.com/gitterHQ/webapp/-/blob/615c78d0b0a314c2c9e4098b8d2ba0471d16961b/modules/templates/lib/safe-json.js
+function safeJson(string) {
+  if (!string) return string;
+  // From http://benalpert.com/2012/08/03/preventing-xss-json.html
+  return string.replace(/<\//g, '<\\/');
+}
```
MadLittleMods, line 8: This is from Gitter.

You can see how we use it in a similar way to inline a JSON payload into the HTML page to generate `window.troupeContext`: https://gitlab.com/gitterHQ/webapp/-/blob/615c78d0b0a314c2c9e4098b8d2ba0471d16961b/modules/templates/lib/hbs-helpers.js#L47-56

PR #28, opened 2022-06-29
Fix archive not responding in big rooms because homeserver times out on `/context` request

MadLittleMods:
Fix archive not responding in big rooms because homeserver times out on `/context` request (too much state).

Add `filter={"lazy_load_members":true}` so that `/context` responds without timing out by returning just the state for the sender of the included event. Otherwise, the homeserver returns all state in the room at that point in time which in big rooms, can be 100k member events that we don't care about anyway. Synapse seems to timeout at about the ~5k state event mark.

## Dev notes

Without `filter={"lazy_load_members":true}`, Synapse can only handle ~4k member events (probably state in general) before `/context` times out before it ever responds in the 180 second window. I'm only looking at the member count as a rough proxy but the number of member events + state will be more.

Loads | Member count | Alias | Matrix.to | API request
--- | --- | --- | --- | ---
:x: | 39k | `#matrix:matrix.org` | [🔗 ](https://matrix.to/#/!OGEhHVWSdvArJzumhm:matrix.org/$xo-tESRNP1Vg1RxLXIfmMeO6dA6-u9XuE2lv6toeKcw?via=matrix.org) | `https://matrix-client.matrix.org/_matrix/client/r0/rooms/!OGEhHVWSdvArJzumhm:matrix.org/context/$xo-tESRNP1Vg1RxLXIfmMeO6dA6-u9XuE2lv6toeKcw?limit=0`
:x: | 31k |  `#openwisp_general:gitter.im` | [🔗 ](https://matrix.to/#/!RBzfoBeqYcCwLAAenz:gitter.im/$YFjmbmH0NRPmfVsPyxhM0jcK4RFR_CdCigtHSgCTLSc?via=matrix.org) |`https://matrix-client.matrix.org/_matrix/client/r0/rooms/!RBzfoBeqYcCwLAAenz:gitter.im/context/$YFjmbmH0NRPmfVsPyxhM0jcK4RFR_CdCigtHSgCTLSc?limit=0`
[...]

PR #29, opened 2022-06-29
Remove unneeded `include_redundant_members` from `/messages` `filter` and test that member state is still visible

MadLittleMods:
Remove unneeded `include_redundant_members` from `/messages` `filter` and add tests to ensure member state is still visible.

Follow-up to https://github.com/matrix-org/matrix-public-archive/pull/28#discussion_r909428366

## Dev notes

Notice the purple user avatar and display name set.

`npm run test-interactive` -> diverse messages test

### DIFF
--- a/test/client-utils.js
+++ b/test/client-utils.js
@@ -34,23 +34,24 @@ async function getTestClientForHs(testMatrixServerUrl) {
     }
   );
 
-  const userId = registerResponse['user_id'];
-  assert(userId);
+  const applicationServiceUserIdOverride = registerResponse['user_id'];
+  assert(applicationServiceUserIdOverride);
 
   return {
     homeserverUrl: testMatrixServerUrl,
     // We use the application service AS token because we need to be able to use
     // the `?ts` timestamp massaging when sending events
     accessToken: matrixAccessToken,
-    userId: userId,
+    userId: applicationServiceUserIdOverride,
+    applicationServiceUserIdOverride,
   };
 }
 
 // Create a public room to test in
 async function createTestRoom(client) {
   let qs = new URLSearchParams();
-  if (client.userId) {
-    qs.append('user_id', client.userId);
+  if (client.applicationServiceUserIdOverride) {
+    qs.append('user_id', client.applicationServiceUserIdOverride);
   }
 
   const createRoomResponse = await fetchEndpointAsJson(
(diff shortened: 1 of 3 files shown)


test/client-utils.js
```diff
@@ -34,23 +34,24 @@ async function getTestClientForHs(testMatrixServerUrl) {
...
+  assert(applicationServiceUserIdOverride);
 
   return {
     homeserverUrl: testMatrixServerUrl,
     // We use the application service AS token because we need to be able to use
     // the `?ts` timestamp massaging when sending events
     accessToken: matrixAccessToken,
-    userId: userId,
+    userId: applicationServiceUserIdOverride,
+    applicationServiceUserIdOverride,
```
MadLittleMods, line 46: Renamed to `applicationServiceUserIdOverride` so it's more obvious what this is used for, what user you want the application service to act on behalf of.

Also added `userId` to reference for the `updateProfile` Matrix API endpoints.

They are separate because we could have a API client which isn't driven by an application service.

test/e2e-tests.js
```diff
@@ -305,8 +322,27 @@ describe('matrix-public-archive', () => {
...
+      assert.match(
+        dom.document.querySelector(`[data-event-id="${imageEventId}"]`).outerHTML,
+        new RegExp(`.*${escapeStringRegexp(displayName)}.*`)
+      );
+
+      // Make sure the user avatar is visible on the message
+      const avatarImageElement = dom.document.querySelector(
+        // FIXME: Use more stable select here instead of `.avatar`,
+        // see https://github.com/vector-im/hydrogen-web/pull/773
+        `[data-event-id="${imageEventId}"] .avatar img`
```
MadLittleMods, line 335: Proposed a change in Hydrogen at https://github.com/vector-im/hydrogen-web/pull/773

test/client-utils.js
```diff
@@ -104,37 +105,44 @@ async function joinRoom({ client, roomId, viaServers }) {
...
 
-  const sendResponse = await fetchEndpointAsJson(
-    urlJoin(
+  let url;
+  if (stateKey) {
+    url = urlJoin(
+      client.homeserverUrl,
+      `/_matrix/client/v3/rooms/${roomId}/state/${eventType}/${stateKey}?${qs.toString()}`
+    );
+  } else {
```
MadLittleMods, line 134: Added support to send state events but ended up using the dedicated `/profile` endpoints in their own `updateProfile` function

PR #31, opened 2022-06-29
Add test for joining a new federated room

MadLittleMods:
Add test for joining a new federated room and making sure the messages are available (homeserver should backfill).

Synapse changes: https://github.com/matrix-org/synapse/pull/13205, https://github.com/matrix-org/synapse/pull/13320

## Dev notes

```
docker build -t matrixdotorg/synapse -f docker/Dockerfile .
```

### DIFF
--- a/server/routes/install-routes.js
+++ b/server/routes/install-routes.js
@@ -137,6 +137,12 @@ function installRoutes(app) {
       // If the hourRange is defined, we force the range to always be 1 hour. If
       // the format isn't correct, redirect to the correct hour range
       if (hourRange && toHour !== fromHour + 1) {
+        // Pass through the query parameters
+        let queryParamterUrlPiece = '';
+        if (req.query) {
+          queryParamterUrlPiece = `?${new URLSearchParams(req.query).toString()}`;
+        }
+
         res.redirect(
           // FIXME: Can we use the matrixPublicArchiveURLCreator here?
           `${urlJoin(
@@ -147,7 +153,7 @@ function installRoutes(app) {
             req.params.mm,
             req.params.dd,
             `${fromHour}-${fromHour + 1}`
-          )}?${new URLSearchParams(req.query).toString()}`
+          )}${queryParamterUrlPiece}`
         );
         return;
       }
--- a/server/routes/install-routes.js
+++ b/server/routes/install-routes.js
@@ -9,6 +9,7 @@ const StatusError = require('../lib/status-error');
 
 const fetchRoomData = require('../fetch-room-data');
 const fetchEventsInRange = require('../fetch-events-in-range');
+const ensureRoomJoined = require('../ensure-room-joined');
 const renderHydrogenToString = require('../render-hydrogen-to-string');
 
 const config = require('../lib/config');
@@ -117,7 +118,8 @@ function installRoutes(app) {
       // the format isn't correct, redirect to the correct hour range
       if (hourRange && toHour !== fromHour + 1) {
         res.redirect(
-          urlJoin(
+          // FIXME: Can we use the matrixPublicArchiveURLCreator here?
+          `${urlJoin(
             basePath,
             roomIdOrAlias,
             'date',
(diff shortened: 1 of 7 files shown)


test/e2e-tests.js
```diff
@@ -106,6 +107,23 @@ describe('matrix-public-archive', () => {
...
+      // FIXME: This doesn't work because `400 Bad Request:
+      // {"errcode":"M_EXCLUSIVE","error":"This user ID is reserved by an
+      // application service."}` and we can't easily remove the AS registration,
+      // register the user, add the AS back, restart Synapse, and continue the
+      // tests.
+      //
+      // Make sure the application service archiver user itself is also
+      // explicitly registered otherwise we run into 404, `Profile was not
+      // found` errors when joining a remote federated room from the archiver
+      // user, see https://github.com/matrix-org/synapse/issues/4778
```
MadLittleMods, line 120: This doesn't work yet, thanks to https://github.com/matrix-org/synapse/issues/4778

And we can't register the username of the application service defined in the registration files.

test/e2e-tests.js
```diff
@@ -379,7 +398,39 @@ describe('matrix-public-archive', () => {
...
+        timestamp: archiveDate.getTime(),
+      });
+
+      archiveUrl = matrixPublicArchiveURLCreator.archiveUrlForDate(hs2RoomId, archiveDate, {
+        // Since hs1 doesn't know about this room on hs2 yet, we have to provide
+        // a via server to ask through.
+        viaServers: ['hs2'],
+      });
+
+      const archivePageHtml = await fetchEndpointAsText(archiveUrl);
```
MadLittleMods, line 421: This doesn't work yet, thanks to https://github.com/matrix-org/synapse/issues/4778 which means the application service user can't join a remote federated room because it will throw a `404` `Profile was not found` unless it's explicitly registered before. 

And we can't register the username of the application service defined in the registration files. So we would have to do some random dance of removing and adding the application service registration and restarting.

test/README.md
```diff
@@ -27,6 +27,19 @@ $ npm run test-interactive
...
+macOS:
+
+```sh
+$ docker ps --all | grep test_hs
+$ docker logs -f --tail 10 matrix_public_archive_test-hs1-1
+$ docker logs -f --tail 10 matrix_public_archive_test-hs2-1
+
+$ docker stop matrix_public_archive_test-hs1-1 matrix_public_archive_test-hs2-1
+$ docker rm matrix_public_archive_test-hs1-1 matrix_public_archive_test-hs2-1
+```
```
MadLittleMods, line 39: Is this actually a macOS vs Windows Docker difference? Or is it just a Docker version difference?

server/routes/install-routes.js
```diff
@@ -117,22 +118,27 @@ function installRoutes(app) {
...
         );
         return;
       }
 
       // TODO: Highlight tile that matches ?at=$xxx
       //const aroundId = req.query.at;
 
+      // We have to wait for the room join to happen first before we can fetch
+      // any of the additional room info or messages.
+      await ensureRoomJoined(matrixAccessToken, roomIdOrAlias, req.query.via);
```
MadLittleMods, line 140: This takes some time even when you're already joined to the room. Is there a faster way to determine whether we are already joined?

Maybe we can just go for it and if we get a `403`/`401`, we can join and re-run the data fetchers.

test/README.md
MadLittleMods: On macOS, just had it create the other numbering of this. idk

shared/lib/url-creator.js
```diff
@@ -1,18 +1,24 @@
...
+const { URLSearchParams } = require('url');
 
 class URLCreator {
   constructor(basePath) {
     this._basePath = basePath;
   }
 
-  archiveUrlForDate(roomId, date) {
+  archiveUrlForDate(roomId, date, { viaServers = [] } = {}) {
+    let qs = new URLSearchParams();
```
MadLittleMods, line 12:
```
URLSearchParams is not a constructor
```

MadLittleMods: This got weirder. It doesn't work on macOS with Chrome Version 103.0.5060.114 but it does work on Windows with the same Chrome version 🤔

test/README.md
MadLittleMods: Removed until I see it again.

test/e2e-tests.js
```diff
@@ -375,7 +392,39 @@ describe('matrix-public-archive', () => {
       );
     });
 
-    it(`can render day back in time from room on remote homeserver we haven't backfilled from`);
+    it.only(`can render day back in time from room on remote homeserver we haven't backfilled from`, async () => {
```
MadLittleMods, line 395: TODO: Remove `it.only`

shared/lib/url-creator.js
MadLittleMods: Works well without importing and using the global `URLSearchParams` ✅

server/routes/install-routes.js
MadLittleMods: Tracked by https://github.com/matrix-org/matrix-public-archive/issues/50

PR #35, opened 2022-07-05
Expose arguments for `renderHydrogenToString` (pure function) so we can reproduce when error occurs while we render Hydrogen

MadLittleMods:
`renderHydrogenToString` is a pure function which means it will give the same output given the same input. This means, that if we give it a certain input and an error occurs, we should be able to reproduce it again if we have the arguments. This PR exposes those arguments so we can investigate what's going wrong.

### DIFF
--- a/server/render-hydrogen-to-string.js
+++ b/server/render-hydrogen-to-string.js
@@ -7,82 +7,92 @@ const { readFile } = require('fs').promises;
 const crypto = require('crypto');
 const { parseHTML } = require('linkedom');
 
+const RethrownError = require('./lib/rethrown-error');
 const config = require('./lib/config');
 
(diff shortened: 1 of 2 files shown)


server/render-hydrogen-to-string.js
```diff
@@ -7,82 +7,92 @@ const { readFile } = require('fs').promises;
...
-  const dom = parseHTML(`
-    <!doctype html>
-    <html>
-      <head></head>
-      <body>
-        <div id="app" class="hydrogen"></div>
-      </body>
-    </html>
-  `);
+    const dom = parseHTML(`
```
MadLittleMods, line 20: It's just indenting everything thanks to the `try/catch` here now.

Issue #34, opened 2022-07-05
App crashes when trying to load particular day of MatrixHQ messages - `TypeError: Cannot read properties of undefined (reading 'storeNames')`


MadLittleMods:
Trying to load MatrixHQ (`https://mpa-deployment/!OGEhHVWSdvArJzumhm:matrix.org/date/2022/07/04`), the app crashes,
```
$ oc logs mpa-statefulset-0 --previous

> matrix-public-archive@0.0.0 start
> node server/server.js

server process.env.NODE_ENV prod
Config is using nodeEnv=prod
/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12481
    let stores = [this._storage.storeNames.timelineEvents];
                                ^

TypeError: Cannot read properties of undefined (reading 'storeNames')
    at TimelineReader.readById (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12481:33)
    at Timeline._getEventFromStorage (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12760:46)
    at Timeline._loadContextEntryNotInTimeline (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12745:35)
    at Timeline._loadContextEntriesWhereNeeded (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12739:14)
    at Timeline.addEntries (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12697:10)
    at mountHydrogen (hydrogen-vm-render-script.js:192:12)
    at hydrogen-vm-render-script.js:336:1
    at Script.runInContext (node:vm:139:12)
    at renderToString (/app/server/render-hydrogen-to-string.js:79:41)
    at async /app/server/routes/install-routes.js:153:34
```

PR #36, opened 2022-07-05
Stop Hydrogen from running in the background after we get our SSR HTML render data

MadLittleMods:
Stop Hydrogen from running in the background after we get our SSR HTML render data.

We now run the Hydrogen render in a `child_process` so we can exit the whole render process. We still use the `vm` to setup the browser-like globals. With a `vm`, everything continues to run even after it returns and there isn't a way to clean up, stop, kill, terminate the vm script or context so we need this extra `child_process` now to clean up. I don't like the complexity necessary for this though. I wish the `vm` API allowed for this use case. The only way to stop a `vm` is the `timeout` and we want to stop as soon as we return.

Fix https://github.com/matrix-org/matrix-public-archive/issues/34

#### Why do we still see an error with no extra context after https://github.com/matrix-org/matrix-public-archive/pull/35?

Matrix public archive is crashing sometimes. I [added some extra logging](https://github.com/matrix-org/matrix-public-archive/pull/35) around the error but it doesn't rethrow the custom error, just the original error in the logs.

You can see the `at renderHydrogenToString (/app/server/render-hydrogen-to-string.js:81:43)` is [within the `try`/`catch`](https://github.com/matrix-org/matrix-public-archive/blob/9c8f980e2a532ba1bb703eb87073412a97153dc5/server/render-hydrogen-to-string.js#L81) that will re-wrap the original error. But we never see that in the logs. Even made sure we are running the version of the code expected by exec'ing into the pod (`oc exec -it mpa-statefulset-0 -- /bin/bash`) to view the source.

```
$ oc logs mpa-statefulset-0 --previous

> matrix-public-archive@0.0.0 start
> node server/server.js

server process.env.NODE_ENV prod
Config is using nodeEnv=prod
/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12481
    let stores = [this._storage.storeNames.timelineEvents];
                                ^

TypeError: Cannot read properties of undefined (reading 'storeNames')
    at TimelineReader.readById (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12481:33)
    at Timeline._getEventFromStorage (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12760:46)
    at Timeline._loadContextEntryNotInTimeline (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12745:35)
    at Timeline._loadContextEntriesWhereNeeded (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12739:14)
    at Timeline.addEntries (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12697:10)
    at mountHydrogen (hydrogen-vm-render-script.js:192:12)
    at hydrogen-vm-render-script.js:336:1
    at Script.runInContext (node:vm:139:12)
    at renderHydrogenToString (/app/server/render-hydrogen-to-string.js:81:43)
    at async /app/server/routes/install-routes.js:153:34
```

See https://github.com/matrix-org/matrix-public-archive/pull/36#discussion_r914116835. It turns out the VM context and script continue running even after we return and it's an error from [async Hydrogen fire and forget code](https://github.com/vector-im/hydrogen-web/blob/28b686dae7df680118507def5cdfff8ba7aee4c7/src/matrix/room/timeline/Timeline.js#L333-L337).

We have to update this to run in a `child_process` so we can exit the process after we get the data we need.


server/render-hydrogen-to-string.js
```diff
@@ -78,21 +78,27 @@ async function renderHydrogenToString({ fromTimestamp, roomData, events, stateEv
...
     });
+    console.log('before runInContext');
     const vmResult = hydrogenRenderScript.runInContext(vmContext);
+    console.log('after runInContext');
     // Wait for everything to render
     // (waiting on the promise returned from `hydrogen-render-script.js`)
     await vmResult;
 
     const documentString = dom.document.body.toString();
+    console.log('got documentString');
```
MadLittleMods, line 89: Does this even make sense? It somehow was able to return the `documentString` but also throw the error

```sh
$ oc logs mpa-statefulset-0 --previous

> matrix-public-archive@0.0.0 start
> node server/server.js

server process.env.NODE_ENV prod
Config is using nodeEnv=prod
before runInContext
after runInContext
got documentString
/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12481
    let stores = [this._storage.storeNames.timelineEvents];
                                ^

TypeError: Cannot read properties of undefined (reading 'storeNames')
    at TimelineReader.readById (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12481:33)
    at Timeline._getEventFromStorage (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12760:46)
    at Timeline._loadContextEntryNotInTimeline (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12745:35)
    at Timeline._loadContextEntriesWhereNeeded (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12739:14)
    at Timeline.addEntries (/app/node_modules/hydrogen-view-sdk/lib-build/hydrogen.cjs.js:12697:10)
    at mountHydrogen (hydrogen-vm-render-script.js:192:12)
    at hydrogen-vm-render-script.js:336:1
    at Script.runInContext (node:vm:139:12)
    at renderHydrogenToString (/app/server/render-hydrogen-to-string.js:82:43)
    at async /app/server/routes/install-routes.js:153:34
```

MadLittleMods: Is the VM continuing to run after returning the result?

MadLittleMods: > Is the VM continuing to run after returning the successful result?

Yes, it's continuing to run. I assumed it closed after the result returned but it runs forever and there isn't a close or stop method so we can do it ourselves.

You can reproduce this by adding some `setTimeout` to `shared/hydrogen-vm-render-script.js`

```js
  setTimeout(() => {
    throw new Error('aewffewaaewfafew');
  }, 5000);

  setInterval(() => {
    console.log('.');
  }, 1000);
```

MadLittleMods: Solved by running all of this inside a `child_process` so we can exit the whole process that does the render after we get the initial SSR HTML.

server/hydrogen-render/render-hydrogen-to-string-fork-script.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+// Called by `child_process` `fork` in `render-hydrogen-to-string.js` so we can
```
MadLittleMods, line 3: The different layers here are getting a bit complex and won't be easy to follow for a new person to the project.

 - `server/hydrogen-render/render-hydrogen-to-string.js`: Runs everything in a `child_process` via `fork`
     - `server/hydrogen-render/render-hydrogen-to-string-fork-script.js`: The script to run in `fork`
         - `server/hydrogen-render/render-hydrogen-to-string-unsafe.js`: Runs everything in a `vm` context to setup the browser-like globals
              - `shared/hydrogen-vm-render-script.js`: Renders Hydrogen to `document.body`

MadLittleMods: I've renamed these with 1-4 to show the usage layers more obviously.

 - `1-render-hydrogen-to-string.js`
 - `2-render-hydrogen-to-string-fork-script.js`
 - `3-render-hydrogen-to-string-unsafe.js`
 - `4-hydrogen-vm-render-script.js`
 
 Better naming? This still isn't great but now at least an easy order is defined for people get some bearing.

server/hydrogen-render/1-render-hydrogen-to-string.js
```diff
@@ -0,0 +1,97 @@
+'use strict';
```
MadLittleMods, line 1: Why the 1-4 naming? see https://github.com/matrix-org/matrix-public-archive/pull/36#discussion_r914239965

server/hydrogen-render/1-render-hydrogen-to-string.js
```diff
@@ -0,0 +1,97 @@
...
+    let data = '';
+    let childErrors = [];
+
+    const controller = new AbortController();
+    const { signal } = controller;
+    // We use a child_process because we want to be able to exit the process after
+    // we receive the SSR results.
+    const child = fork(
+      require.resolve('./2-render-hydrogen-to-string-fork-script'),
+      [JSON.stringify(options)],
```
MadLittleMods, line 26: Passing the options via argv doesn't work with big payloads, `Error: spawn E2BIG`

Solved in https://github.com/matrix-org/matrix-public-archive/pull/37

server/hydrogen-render/render-hydrogen-to-string-fork-script.js
MadLittleMods: Updated in https://github.com/matrix-org/matrix-public-archive/pull/64 to separate the concerns of `child_process` runner and Hydrogen render

PR #38, opened 2022-07-06
Make sure we finish sending the HTML payload before we exit the process

MadLittleMods:
Make sure we finish sending the HTML payload before we exit the process.

I encountered a page which responded successfully but all of the Hydrogen HTML was missing. It just had the boilerplate around it.

What I am guessing happened is that since `process.send` is async, with a sufficiently large
payload and race condition, `process.exit(0)` was being called before it finished sending.

Related:
 - https://stackoverflow.com/questions/34627546/process-send-is-sync-async-on-nix-windows
 - https://github.com/nodejs/node/commit/56d9584a0ead78874ca9d4de2e55b41c4056e502
 - https://github.com/nodejs/node/issues/6767

### DIFF
--- a/server/hydrogen-render/1-render-hydrogen-to-string.js
+++ b/server/hydrogen-render/1-render-hydrogen-to-string.js
@@ -6,7 +6,7 @@
 
 const fork = require('child_process').fork;
 
-const { assert } = require('console');
+const assert = require('assert');
 const RethrownError = require('../lib/rethrown-error');
 
 // The render should be fast. If it's taking more than 5 seconds, something has
--- a/server/hydrogen-render/1-render-hydrogen-to-string.js
+++ b/server/hydrogen-render/1-render-hydrogen-to-string.js
@@ -6,6 +6,7 @@
 
 const fork = require('child_process').fork;
 
+const { assert } = require('console');
 const RethrownError = require('../lib/rethrown-error');
 
 // The render should be fast. If it's taking more than 5 seconds, something has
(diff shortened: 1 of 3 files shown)


server/hydrogen-render/1-render-hydrogen-to-string.js
```diff
@@ -84,6 +89,11 @@ async function renderHydrogenToString(options) {
       });
     });
 
+    assert(
+      data,
+      `No HTML sent from child process to render Hydrogen. Any child errors? (${childErrors.length})`
```
MadLittleMods, line 94: We don't expect any `childErrors` if this is ever encountered but is just a sanity check. We don't even give the relevant info if there was `childErrors` 🤷

server/hydrogen-render/1-render-hydrogen-to-string.js
```diff
@@ -60,12 +61,16 @@ async function renderHydrogenToString(options) {
...
+              childErrors.length
+            } errors but we really always expect 1 error)\n${childErrors
+              .map((childError, index) => ` ${index}. ${childError.message} ${childError.stack}`)
+              .join('\n')}`;
           }
 
           const error = new RethrownError(
             `Child process failed with exit code ${exitCode}${extraErrorsMessage}`,
-            childErrors[0]
+            childErrors[0] || new Error('No child errors')
```
MadLittleMods, line 73: This is possible if the process exits for some other random reason. Exit code can even by `null`.

server/hydrogen-render/1-render-hydrogen-to-string.js
```diff
@@ -60,12 +61,16 @@ async function renderHydrogenToString(options) {
         } else {
           let extraErrorsMessage = '';
           if (childErrors.length > 1) {
-            extraErrorsMessage = ` (somehow we saw ${childErrors.length} errors but we really always expect 1 error)`;
+            extraErrorsMessage = ` (somehow we saw ${
+              childErrors.length
+            } errors but we really always expect 1 error)\n${childErrors
+              .map((childError, index) => ` ${index}. ${childError.message} ${childError.stack}`)
+              .join('\n')}`;
```
MadLittleMods, line 68: This will be ugly but I thought that we should give the actually relevant info if this ever happens.

server/hydrogen-render/2-render-hydrogen-to-string-fork-script.js
```diff
@@ -13,12 +15,31 @@ process.on('message', async (options) => {
   try {
     const resultantHtml = await _renderHydrogenToStringUnsafe(options);
 
-    // Send back the data we need
-    process.send({
-      data: resultantHtml,
+    assert(resultantHtml, `No HTML returned from _renderHydrogenToStringUnsafe.`);
+
+    // Send back the data we need to the parent.
+    await new Promise((resolve, reject) => {
```
MadLittleMods, line 21: This is wrapped in a promise because `process.send(...)` is async and only has a `callback` option. And we want any error from the callback to be caught by our general `try`/`catch` out here

Issue #49, opened 2022-08-29
Running the Hydrogen render process in `child_process` is slow


MadLittleMods:
*Spawning from https://github.com/matrix-org/matrix-public-archive/pull/44 which added more tracing and made it obvious how much slow `child_process` is over the base `vm` `script.runInNewContext(...)`*

---

## Soo, how slow is running in the `child_process`?

In https://github.com/matrix-org/matrix-public-archive/pull/36, we started wrapping our whole rendering Hydrogen `vm` flow in a `child_process` because using a Node.js `vm` context doesn't have a way to stop the process (it just keeps running). With the `child_process` we can kill it when we choose after we get the HTML result.

Now with this timing from https://github.com/matrix-org/matrix-public-archive/pull/44, we can also get some insight into how bad of a performance hit it is to run the Hydrogen render in the `child_process` is. It seems to be a lot slower: ~50ms vs 360ms.

```js
// Use the `vm` directly to see how fast it is
const _renderHydrogenToStringUnsafe = require('../hydrogen-render/3-render-hydrogen-to-string-unsafe');
const hydrogenHtmlOutput = await _renderHydrogenToStringUnsafe({ /* renderData */ });
```

![](https://user-images.githubusercontent.com/558581/180586026-ff6c653e-a54d-4cf4-abc8-a8c51971aad5.png)

#### Should we switch away from `child_process`?

Maybe.

We could use the `vm` `script.runInNewContext(...)` `timeout` option and just accept that it will be living for 5 seconds (or whatever we timeout we choose) before it dies. We already choose to kill the `child_process` after 5 seconds if it takes too long since we expect the Hydrogen render part to be fast.

Or maybe abuse the `vm` `script.runInNewContext(...)` `breakOnSigint` option and send a `SIGINT` to kill it. Need to be careful not to kill our main process.

Is it possible just to reset and render Hydrogen over and over with the same VM context? We could do a similar thing and keep the `child_process` alive and share for all renders.

PR #43, opened 2022-07-20
Add test to make sure the archive doesn't fail when event for event relation is missing and not included in list of provided events

MadLittleMods:
Add test to make sure the archive doesn't fail when event for event relation is missing and not included in list of provided events. Like if someone is replying to an event that was from long ago out of our range.

In the case of missing relations, Hydrogen does `_loadContextEntryNotInTimeline` because it can't find the event locally which throws an `uncaughtException`. Before https://github.com/matrix-org/matrix-public-archive/pull/51, the `uncaughtException` killed the Hydrogen `child_process` before it could pass back the HTML. Now this PR mainly just adds a test to make sure it works.
```
TypeError: Cannot read properties of undefined (reading 'storeNames')
    at TimelineReader.readById (C:\Users\MLM\Documents\GitHub\element\hydrogen-web\target\lib-build\hydrogen.cjs.js:12483:33)
    at Timeline._getEventFromStorage (C:\Users\MLM\Documents\GitHub\element\hydrogen-web\target\lib-build\hydrogen.cjs.js:12762:46)
    at Timeline._loadContextEntryNotInTimeline (C:\Users\MLM\Documents\GitHub\element\hydrogen-web\target\lib-build\hydrogen.cjs.js:12747:35)
    at Timeline._loadContextEntriesWhereNeeded (C:\Users\MLM\Documents\GitHub\element\hydrogen-web\target\lib-build\hydrogen.cjs.js:12741:14)
    at Timeline.addEntries (C:\Users\MLM\Documents\GitHub\element\hydrogen-web\target\lib-build\hydrogen.cjs.js:12699:10)
    at mountHydrogen (4-hydrogen-vm-render-script.js:204:12)
    at 4-hydrogen-vm-render-script.js:353:1
    at Script.runInContext (node:vm:139:12)
    at _renderHydrogenToStringUnsafe (C:\Users\MLM\Documents\GitHub\element\matrix-public-archive\server\hydrogen-render\3-render-hydrogen-to-string-unsafe.js:102:41)
    at async process.<anonymous> (C:\Users\MLM\Documents\GitHub\element\matrix-public-archive\server\hydrogen-render\2-render-hydrogen-to-string-fork-script.js:18:27)
```

## Other fix: More robust child process error catching

**Update:** These updates were split out to https://github.com/matrix-org/matrix-public-archive/pull/51

Also fix underlying error being obscured:

Before:

```
[...]

PR #51, opened 2022-08-30
Make the `child_process` error catching more robust

MadLittleMods:
Make the `child_process` error catching more robust

Split off from https://github.com/matrix-org/matrix-public-archive/pull/43

Listen to `process.on('uncaughtException', ...)` and handle the async errors ourselves so it no longer fails the child process.

And if the process does exit with status code 1 (error), we have those underlying errors serialized and shown.

---

In the future, maybe we want to switch away from `child_process`, https://github.com/matrix-org/matrix-public-archive/issues/49, which would also be nice to remove all of this complexity.

### DIFF
--- a/server/hydrogen-render/1-render-hydrogen-to-string.js
+++ b/server/hydrogen-render/1-render-hydrogen-to-string.js
@@ -10,11 +10,20 @@ const assert = require('assert');
 const RethrownError = require('../lib/rethrown-error');
 const { traceFunction } = require('../tracing/trace-utilities');
 
+const config = require('../lib/config');
+const logOutputFromChildProcesses = config.get('logOutputFromChildProcesses');
+
 // The render should be fast. If it's taking more than 5 seconds, something has
 // gone really wrong.
 const RENDER_TIMEOUT = 5000;
 
-async function renderHydrogenToString(options) {
+if (!logOutputFromChildProcesses) {
+  console.warn(
+    `Silencing logs from child processes (config.logOutputFromChildProcesses = ${logOutputFromChildProcesses})`
+  );
+}
+
+async function renderHydrogenToString(renderOptions) {
   try {
     let data = '';
     let childErrors = [];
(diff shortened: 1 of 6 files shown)


server/hydrogen-render/1-render-hydrogen-to-string.js
```diff
@@ -65,13 +94,20 @@ async function renderHydrogenToString(options) {
             extraErrorsMessage = ` (somehow we saw ${
               childErrors.length
             } errors but we really always expect 1 error)\n${childErrors
-              .map((childError, index) => ` ${index}. ${childError.message} ${childError.stack}`)
+              .map((childError, index) => `${index}. ${childError.stack}`)
```
MadLittleMods, line 97: Prevent the error message from being duplicated. It's already included in the `stack`.

server/lib/rethrown-error.js
```diff
@@ -40,7 +40,7 @@ class RethrownError extends ExtendedError {
     const messageLines = (this.message.match(/\n/g) || []).length + 1;
 
     const indentedOriginalError = error.stack
-      .split('\n')
+      .split(/\r?\n/)
```
MadLittleMods, line 43: Fix multiple newlines appearing from child errors we emit from `stderr` output (Windows problem).

server/hydrogen-render/2-render-hydrogen-to-string-fork-script.js
```diff
@@ -6,14 +6,60 @@
...
+}
+
+// We don't exit the process after encountering one of these because maybe it
+// doesn't matter to the main render process in Hydrogen.
+//
+// If we don't listen for these events, the child will exit with status code 1
+// (error) when they occur.
+process.on('uncaughtException', async (err /*, origin*/) => {
+  await serializeError(new RethrownError('uncaughtException in child process', err));
+});
```
MadLittleMods, line 51: The really nice fix in this PR is listening to `uncaughtException` here ^

We keep track of the errors to display if we fail to get output from Hydrogen so it's easier to debug.

But these don't fail the process because they're out of band from the main Hydrogen rendering process and don't necessarily mean the Hydrogen render failed. Like `localStorage` and `IndexDB` errors.

PR #62, opened 2022-09-02
Better child process error handling

MadLittleMods:
Better child process error handling

### DIFF
--- a/server/hydrogen-render/1-render-hydrogen-to-string.js
+++ b/server/hydrogen-render/1-render-hydrogen-to-string.js
@@ -23,10 +23,41 @@ if (!logOutputFromChildProcesses) {
   );
 }
 
+function assembleErrorAfterChildExitsWithErrors(exitCode, childErrors) {
+  assert(childErrors);
+
+  let extraErrorsMessage = '';
+  if (childErrors.length > 1) {
+    extraErrorsMessage = ` (somehow we saw ${
+      childErrors.length
+    } errors but we really always expect 1 error)\n${childErrors
+      .map((childError, index) => `${index}. ${childError.stack}`)
+      .join('\n')}`;
+  }
+
+  let childErrorToDisplay;
+  if (childErrors.length === 0) {
+    childErrorToDisplay = new Error('No child errors');
+    childErrorToDisplay.stack = 'No child errors';
+  } else if (childErrors.length === 1) {
+    childErrorToDisplay = childErrors[0];
+  } else {
+    childErrorToDisplay = new Error('Multiple child errors listed above ^');
+  }
+
+  const childErrorSummary = new RethrownError(
+    `Child process exited with code ${exitCode}${extraErrorsMessage}`,
+    childErrorToDisplay
+  );
+
+  return childErrorSummary;
+}
+
 async function renderHydrogenToString(renderOptions) {
+  let abortTimeoutId;
   try {
-    let data = '';
     let childErrors = [];
+    let childExitCode = '(not set yet)';
 
     const controller = new AbortController();
     const { signal } = controller;
@@ -60,11 +91,12 @@ async function renderHydrogenToString(renderOptions) {
     child.send(renderOptions);
 
     // Stops the child process if it takes too long
-    setTimeout(() => {
+    abortTimeoutId = setTimeout(() => {
       controller.abort();
     }, RENDER_TIMEOUT);
 
-    await new Promise((resolve, reject) => {
+    const returnedData = await new Promise((resolve, reject) => {
+      let data = '';
       // Collect the data passed back by the child
       child.on('message', function (result) {
         if (result.error) {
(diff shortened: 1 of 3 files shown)


server/hydrogen-render/1-render-hydrogen-to-string.js
```diff
@@ -85,60 +117,58 @@ async function renderHydrogenToString(renderOptions) {
...
     throw new RethrownError(
       `Failed to render Hydrogen to string. In order to reproduce, feed in these arguments into \`renderHydrogenToString(...)\`:\n    renderToString arguments: ${JSON.stringify(
         renderOptions
       )}`,
       err
     );
+  } finally {
+    // We don't have to add a undefined/null check here because `clearTimeout`
+    // works with any value you give it and doesn't throw an error.
+    clearTimeout(abortTimeoutId);
```
MadLittleMods, line 171: This wasn't strictly necessary but it's good practice to stop the timeout callback from aborting nothing later.

server/hydrogen-render/1-render-hydrogen-to-string.js
```diff
@@ -85,60 +117,58 @@ async function renderHydrogenToString(renderOptions) {
...
-            err
+          const childErrorSummary = assembleErrorAfterChildExitsWithErrors(
+            childExitCode,
+            childErrors
+          );
+          reject(
+            new RethrownError(
+              `Timed out while rendering Hydrogen to string so we aborted the child process after ${RENDER_TIMEOUT}ms. Any child errors? (${childErrors.length})`,
+              childErrorSummary
+            )
```
MadLittleMods, line 144: When we timeout rendering, show any child errors that might give us more context for what went wrong.

server/hydrogen-render/1-render-hydrogen-to-string.js
```diff
@@ -85,60 +117,58 @@ async function renderHydrogenToString(renderOptions) {
...
         if (err.name === 'AbortError') {
-          throw new RethrownError(
-            `Timed out while rendering Hydrogen to string so we aborted the child process after ${RENDER_TIMEOUT}ms`,
-            err
+          const childErrorSummary = assembleErrorAfterChildExitsWithErrors(
+            childExitCode,
+            childErrors
+          );
+          reject(
+            new RethrownError(
```
MadLittleMods, line 141: Instead of throwing this error, we now reject with it which seems a lot more proper.

I'm not sure how it was able to bubble up to our try/catch before when this is in a callback 🤔 (it does do that though).

server/hydrogen-render/2-render-hydrogen-to-string-fork-script.js
```diff
@@ -88,10 +89,10 @@ process.on('message', async (renderOptions) => {
...
   } catch (err) {
     // We need to wait for the error to completely send to the parent
-    // process before we throw the error again and exit the process.
+    // process before we exit the process.
     await serializeError(err);
 
-    // Throw the error so the process fails and exits
-    throw err;
+    // Fail the process and exit
+    process.exit(1);
```
MadLittleMods, line 96: This is an important change! The previous code didn't actually fail and exit the process for us and it's unclear to me if it ever did.

This works to show us a `TypeError: ...` immediately for example instead of having the render process timeout and mask the error.

server/lib/rethrown-error.js
```diff
@@ -32,7 +32,7 @@ class RethrownError extends ExtendedError {
     super(message);
     if (!error) throw new Error('RethrownError requires a message and error');
     this.original = error;
-    this.newStack = this.stack;
+    this.originalStack = this.stack;
```
MadLittleMods, line 35: This property isn't ever used but seemed like the proper name now 🤷

server/hydrogen-render/1-render-hydrogen-to-string.js
```diff
@@ -85,60 +117,58 @@ async function renderHydrogenToString(renderOptions) {
...
-    assert(
-      data,
-      `No HTML sent from child process to render Hydrogen. Any child errors? (${childErrors.length})`
-    );
+    if (!returnedData) {
+      const childErrorSummary = assembleErrorAfterChildExitsWithErrors(childExitCode, childErrors);
+      throw new RethrownError(
+        `No HTML sent from child process to render Hydrogen. Any child errors? (${childErrors.length})`,
+        childErrorSummary
+      );
```
MadLittleMods, line 157: Also good to give more context if this situation ever happens.

server/hydrogen-render/1-render-hydrogen-to-string.js
```diff
@@ -23,10 +23,41 @@ if (!logOutputFromChildProcesses) {
...
+
+  return childErrorSummary;
+}
+
 async function renderHydrogenToString(renderOptions) {
+  let abortTimeoutId;
   try {
-    let data = '';
     let childErrors = [];
+    let childExitCode = '(not set yet)';
```
MadLittleMods, line 60: Whenever we log `childErrors`, we now also show the `childExitCode` and this is a useful default over `null` or `undefined` which can be a head scratcher whether the code is even doing the right thing.

server/hydrogen-render/1-render-hydrogen-to-string.js
```diff
@@ -23,10 +23,48 @@ if (!logOutputFromChildProcesses) {
...
+    childErrorToDisplay.stack = childErrorToDisplay.message;
+  }
+
+  const childErrorSummary = new RethrownError(
+    `Child process exited with code ${exitCode}${extraErrorsMessage}`,
+    childErrorToDisplay
+  );
+
+  return childErrorSummary;
+}
```
MadLittleMods, line 61: Generalizing this so we can re-use it in various scenarios.

Issue #67, opened 2022-09-08
Add middleware to prevent clickjacking exploits


MadLittleMods:
Add middleware to prevent clickjacking exploits

https://gitlab.com/gitterHQ/webapp/-/blob/1116cc549d6d5d22e668c914664d0c78731f436e/server/web/middlewares/prevent-clickjacking.js

```js
// Don't allow others to iframe embed which can lead to clickjacking
function preventClickjackingMiddleware(req, res, next) {
  res.set('X-Frame-Options', 'DENY');

  next();
}
```

PR #68, opened 2022-09-09
Add clickjacking prevention middleware

MadLittleMods:
Add clickjacking prevention middleware

Fix https://github.com/matrix-org/matrix-public-archive/issues/67

### DIFF
--- a/server/routes/install-routes.js
+++ b/server/routes/install-routes.js
@@ -6,9 +6,11 @@ const asyncHandler = require('../lib/express-async-handler');
 
 const { handleTracingMiddleware } = require('../tracing/tracing-middleware');
 const getVersionTags = require('../lib/get-version-tags');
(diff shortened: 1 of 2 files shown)


server/routes/install-routes.js
```diff
@@ -6,9 +6,11 @@ const asyncHandler = require('../lib/express-async-handler');
 
 const { handleTracingMiddleware } = require('../tracing/tracing-middleware');
 const getVersionTags = require('../lib/get-version-tags');
+const preventClickjackingMiddleware = require('./prevent-clickjacking-middleware');
 
 function installRoutes(app) {
   app.use(handleTracingMiddleware);
+  app.use(preventClickjackingMiddleware);
```
MadLittleMods, line 13: Tested and works to prevent embedding the archive in an iframe ✅:

```html
<iframe src="http://localhost:3050/"></iframe>
```

Issue #46, opened 2022-08-01
Add jump to next and previous activity in room


MadLittleMods:
Rooms can be gappy and inactive for days at a time. Instead of having to manually navigate day by day in the archive trying to find the previous message, it would be better if you could just jump to where the next event is.

Under the calendar in the sidebar we can add, `Jump to next activity in room`, `Jump to previous activity in room` (better way to phrase?)

I was also thinking of adding it to the timeline and especially showcase it in the empty view when there are no events for the day.

If there is no next/previous activity, consider following tombstone and predecessor history, https://github.com/matrix-org/matrix-public-archive/issues/59

## Dev notes

We can use `/timestamp_to_event` again to find the next/previous event in the room.

For the archive, we would have the `/:roomId/date/2022/02/08/next`  endpoint which would redirect you to the next date with activity (same with `/previous`)

Could be a more general endpoint: `/:roomId/jump?ts=xxx&dir=[f|b]`

PR #71, opened 2022-09-15
Show surrounding messages for a full screen of content

MadLittleMods:
#### 1. Full screen of content

Add surrounding messages to the given messages so we have a full screen of content to make it feel lively even in quiet rooms

As you scroll around the timeline across different days, the date changes in the URL, calendar, etc

#### 2. Day summary item and jump to next activity

Add summary item to the bottom of the timeline that explains if we couldn't find any messages in the specific day requested 

No events on day | No events at all
--- | ---
 | 

Also allows you to the jump to the next activity in the room. Adds `/:roomId/jump?ts=xxx&dir=[f|b]` to facilitate this.

Part of https://github.com/matrix-org/matrix-public-archive/issues/46

#### 3. Add developer options moda

Add developer options modal which is linked from the bottom of the right-panel
    - Adds an option so you can debug the `IntersectionObserver` and how it's selecting the active day from the top-edge of the scroll viewport.
 - In the future, this will also include a nice little visualization of the backend timing traces

### One person's link will not look like the other once they start scrolling

Because we change the URL as you scroll around. Someone may link a page expecting all the same messages to be there but it will actually only be the current day they're scrolled to plus whatever is above it.

This would be mediated if we could also show messages after the selected day, but we can't (see the next point).

### We can't show messages after the selected day

With `/messages`, we can only pick an event and paginate backwards because paginating forwards does not work properly with `/backfill`. And `/context` doesn't `/backfill` at all.

Because we can't get messages after the selected day, if the day doesn't have enough messages to fill up the screen, the URL will immediately change to a day before and there is no way to scroll low enough to select the day again except for the calendar.

This is now tracked separately as https://github.com/matrix-org/matrix-public-archive/issues/72

### Scroll to the start of the day

Currently, the timeline always scrolls to the bottom on page-load but it would be nice to be scrolled to the beginning of the day. This is tracked by https://github.com/matrix-org/matrix-public-archive/issues/73

#### `IntersectionObserver` notes

 - https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API
 - https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver

Only showing intersections on the top edge:

 - https://stackoverflow.com/a/54874286/796832
 - https://css-tricks.com/an-explanation-of-how-the-intersection-observer-watches/#aa-creating-a-position-sticky-event
 
---

The gif shows off how scrolling around changes the calendar day. Also shows off how we're using an `IntersectionObserver` to select the active day at the top edge of the scroll root:

 
#### Scroll to event
 
A little change to avoid extra resizes messing with the scroll position,

[Hydrogen `src/platform/web/ui/session/room/TimelineView.ts#L94-L99`](https://github.com/vector-im/hydrogen-web/blob/0e40258404f09221a04b23537b1a7b8a258321f7/src/platform/web/ui/session/room/TimelineView.ts#L94-L99)
 
 ```js
if (typeof ResizeObserver === "function") {
    let inititialResize = true;
    this.resizeObserver = new ResizeObserver((entries) => {
        console.log('resizeObserver', inititialResize, entries);
        // The `ResizeObserver` is invoked when initialized but we
        // already do the initial `restoreScrollPosition` in the
        // `requestAnimationFrame(...)` above so we can ignore this
        // extra call.
        //
        // Note: There can still be layout shifts from the webfonts
        // loading in.
        if(!inititialResize) {
            this.restoreScrollPosition();
        }
        inititialResize = false;
    });
    this.resizeObserver.observe(root);
}
```

### `nconf` config reset

Seems to work as expected but I am not sure how it works. How does [`nconf`](https://github.com/indexzero/nconf) know when to "snapshot" some state and reset back to it? There isn't any documentation on `reset` and the [code](https://github.com/indexzero/nconf/blob/2db6af48d36fffc563c6aeb24f4942726e47cbd2/lib/nconf/stores/memory.js#L240-L252) isn't illuminating.

```js
console.log('Num config keys at start', Object.keys(config.get()).length);
console.log('config', config.get('archiveMessageLimit'));

config.set('archiveMessageLimit', 9999);
console.log('config', config.get('archiveMessageLimit'));

config.reset();

console.log('config', config.get('archiveMessageLimit'));
console.log('Num config keys at end', Object.keys(config.get()).length);
```

```
Num config keys at start 125
config 500
config 9999
config 500
Num config keys at end 125
```

---

Related issues: 

 - https://github.com/indexzero/nconf/issues/93#issuecomment-1064470817
 - https://github.com/indexzero/nconf/issues/145

 
## Todo

 - [ ] ~~Start the scroll at the beginning of a day~~
    - Opting for the simple solution of leaving it as-is (scroll to the bottom)
    - Tracking this nicety in https://github.com/matrix-org/matrix-public-archive/issues/73
 - [ ] ~~Optional for this PR: Add date headers to Hydrogen~~
    - https://github.com/vector-im/hydrogen-web/pull/764
 - [x] Even though we return events outside of the day, check to make sure we don't have too many events from the active day to determine whether we need to redirect to hour segments
 - [x] Add a way to toggle developer settings on like the `IntersectionObserver` highlight debugging styles
    - Can also include the tracing span timing in here
  - [x] Make sure opening a lightbox, will change the URL for the day with the lightbox so the link still brings you to the lightbox in question when you open it later
  - [x] Make sure the message in the disabled composer, "You're viewing an archive of events from 2022-02-08", gets updated as you scroll around
  - [x] Add placeholder when there are no events for the active day. Otherwise, it's weird to navigate to a specific day, only see events from a previous day and have the active day changed based on the scroll of other events (you question whether you went to the right place or if the archive messed up)
     - Maybe this could also solve when there aren't enough events for a given day of scroll
 - [x] Add tests for different day summary scenarios
 - [x] Add tests for day with too many messages
 - [x] Add tests for too many messages but they are from surrounding days so it's ok
 - [x] Update `hydrogen-view-sdk` again with latest [scratch changes](https://github.com/vector-im/hydrogen-web/pull/653)
    - https://github.com/matrix-org/matrix-public-archive/pull/71#discussion_r973546587

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "@opentelemetry/semantic-conventions": "^1.3.1",
     "dompurify": "^2.3.9",
     "express": "^4.17.2",
-    "hydrogen-view-sdk": "npm:@mlm/hydrogen-view-sdk@^0.16.0-scratch",
+    "hydrogen-view-sdk": "npm:@mlm/hydrogen-view-sdk@^0.17.0-scratch",
     "json5": "^2.2.1",
     "linkedom": "^0.14.1",
     "matrix-public-archive-shared": "file:./shared/",
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "@opentelemetry/semantic-conventions": "^1.3.1",
     "dompurify": "^2.3.9",
     "express": "^4.17.2",
-    "hydrogen-view-sdk": "npm:@mlm/hydrogen-view-sdk@^0.17.0-scratch",
+    "hydrogen-view-sdk": "npm:@mlm/hydrogen-view-sdk@^0.18.0-scratch",
     "json5": "^2.2.1",
     "linkedom": "^0.14.1",
     "matrix-public-archive-shared": "file:./shared/",
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "@opentelemetry/semantic-conventions": "^1.3.1",
         "dompurify": "^2.3.9",
         "express": "^4.17.2",
-        "hydrogen-view-sdk": "npm:@mlm/hydrogen-view-sdk@^0.17.0-scratch",
+        "hydrogen-view-sdk": "npm:@mlm/hydrogen-view-sdk@^0.18.0-scratch",
         "json5": "^2.2.1",
         "linkedom": "^0.14.1",
         "matrix-public-archive-shared": "file:./shared/",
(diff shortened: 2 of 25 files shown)


package.json
```diff
@@ -47,7 +47,7 @@
     "@opentelemetry/semantic-conventions": "^1.3.1",
     "dompurify": "^2.3.9",
     "express": "^4.17.2",
-    "hydrogen-view-sdk": "npm:@mlm/hydrogen-view-sdk@^0.16.0-scratch",
+    "hydrogen-view-sdk": "npm:@mlm/hydrogen-view-sdk@^0.17.0-scratch",
```
MadLittleMods, line 50: TODO: Update the dependency again

PR #81, opened 2022-10-18
Add `Content-Security-Policy` (CSP)

MadLittleMods:
Add `Content-Security-Policy` (CSP) that restricts the page to just what it is expected to do.

This helps limit the damage that can be done by any XSS attack.

Fix https://github.com/matrix-org/internal-config/issues/1341

### Todo

 - [x] Add `nonce` support to the timeout error page (`timeout-middleware`)

### DIFF
--- a/server/routes/content-security-policy-middleware.js
+++ b/server/routes/content-security-policy-middleware.js
@@ -32,6 +32,10 @@ function contentSecurityPolicyMiddleware(req, res, next) {
     `media-src 'self' ${matrixServerUrl};`,
     // Only allow the room directory search form to submit to ourself
     `form-action 'self';`,
+    // We have no need ourself to embed in an iframe. And we shouldn't allow others to
+    // iframe embed which can lead to clickjacking.  We also have the
+    // `prevent-clickjacking-middleware` to cover this.
+    `frame-ancestors 'none';`,
   ];
 
   res.set('Content-Security-Policy', directives.join(' '));
--- a/server/routes/content-security-policy-middleware.js
+++ b/server/routes/content-security-policy-middleware.js
@@ -42,7 +42,7 @@ function contentSecurityPolicyMiddleware(req, res, next) {
 
   res.set('Content-Security-Policy', directives.join(' '));
 
-  // Make this available for down-stream routes to use
+  // Make this available for down-stream routes to reference and use
   res.locals.cspNonce = nonce;
 
   next();
--- a/server/routes/content-security-policy-middleware.js
+++ b/server/routes/content-security-policy-middleware.js
@@ -36,6 +36,8 @@ function contentSecurityPolicyMiddleware(req, res, next) {
     // iframe embed which can lead to clickjacking.  We also have the
     // `prevent-clickjacking-middleware` to cover this.
     `frame-ancestors 'none';`,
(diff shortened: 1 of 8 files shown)


server/routes/content-security-policy-middleware.js
```diff
@@ -0,0 +1,45 @@
...
+    // Only only <script> and <style> tags that have the nonce provided
+    //
+    // To ensure compatibility with very old browser versions (4+ years), we add
+    // 'unsafe-inline' as a fallback. All recent browsers will ignore 'unsafe-inline' if
+    // a CSP nonce or hash is present. (via
+    // https://web.dev/strict-csp/#step-4-add-fallbacks-to-support-safari-and-older-browsers)
+    `script-src 'nonce-${nonce}' 'strict-dynamic' https: 'unsafe-inline';`,
+    `style-src 'nonce-${nonce}' 'strict-dynamic' https: 'unsafe-inline';`,
+    // Hydrogen uses a bunch of inline styles
+    `style-src-attr 'unsafe-inline';`,
```
MadLittleMods, line 26: `style-src-attr` doesn't really have [support in Firefox](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src-attr#browser_compatibility) yet (only supported in Nightly).

How can we craft a directive to only allow `<style nonce="...">` and inline styles? This seems hard since all recent browsers will ignore 'unsafe-inline' if a CSP nonce or hash is present.

MadLittleMods: We could use `style-src self 'unsafe-inline';` but that's not *better* than a `nonce`. We might have to settle for that 🤔

server/routes/content-security-policy-middleware.js
```diff
@@ -0,0 +1,45 @@
...
+    // We only need to load fonts from ourself
+    `font-src 'self';`,
+    // We only need to be able to load images/media from ourself and the homeserver media repo
+    `img-src 'self' ${matrixServerUrl};`,
+    `media-src 'self' ${matrixServerUrl};`,
+    // It seems like we would also need the `form-action` directive but since we don't
+    // use any `action` on the room directory search form, it seems to work without
+    // needing to allow it here.
```
MadLittleMods, line 34: I'd be curious to know why our search `<form method="GET">` works without setting a `form-action` directive. There is no `action` on the form but I'd assume the default of the current page would factor in 🤔

MadLittleMods: Ahh, `default-src` is only for other `-src` (fetch) directives, https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/default-src

So it would be good to define `form-action` here ⏩

MadLittleMods: Added ✅

server/routes/content-security-policy-middleware.js
dkasak: I think it's reasonable to settle for that, given constraints.

dkasak (approved):
Looks reasonable to me.

MadLittleMods:
Thanks for taking a look @dkasak 🐅

Issue #80, opened 2022-10-14
Searching for a Null Byte Triggers a Server Error


davidegirardi:
Visiting this URL triggers a 500: `/?search=%00`

It _could_ be an issue in `node-fetch` but I am not familiar enough with the code to be sure.

MadLittleMods:
@davidegirardi I've updated the error message to be a bit more understandable: https://github.com/matrix-org/matrix-public-archive/pull/84

But you should also raise this in the Synapse issue tracker since that input to `/_matrix/client/v3/publicRooms`  causes a 500 error:

```
2022-10-18 03:55:45,082 - synapse.http.server - 123 - ERROR - POST-51351 - Failed handle request via 'PublicRoomListRestServlet': <XForwardedForRequest at 0x10f50d4c0 method='POST' uri='/_matrix/client/v3/publicRooms?' clientproto='HTTP/1.1' site='8008'>
Traceback (most recent call last):
  File "synapse/synapse/http/server.py", line 306, in _async_render_wrapper
    callback_return = await self._async_render(request)
  File "synapse/synapse/http/server.py", line 512, in _async_render
    callback_return = await raw_callback_return
  File "synapse/synapse/rest/client/room.py", line 544, in on_POST
    data = await handler.get_local_public_room_list(
  File "synapse/synapse/handlers/room_list.py", line 103, in get_local_public_room_list
    return await self._get_public_room_list(
  File "synapse/synapse/handlers/room_list.py", line 164, in _get_public_room_list
    results = await self.store.get_largest_public_rooms(
  File "synapse/synapse/storage/databases/main/room.py", line 472, in get_largest_public_rooms
    ret_val = await self.db_pool.runInteraction(
  File "synapse/synapse/storage/database.py", line 881, in runInteraction
    return await delay_cancellation(_runInteraction())
  File "synapse/.venv/lib/python3.9/site-packages/twisted/internet/defer.py", line 1656, in _inlineCallbacks
    result = current_context.run(
  File "synapse/.venv/lib/python3.9/site-packages/twisted/python/failure.py", line 514, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "synapse/synapse/storage/database.py", line 848, in _runInteraction
    result = await self.runWithConnection(
  File "synapse/synapse/storage/database.py", line 976, in runWithConnection
    return await make_deferred_yieldable(
[...]

PR #84, opened 2022-10-18
Throw more understandable error when we fail to fetch from the homeserver room directory

MadLittleMods:
Throw more understandable error when we fail to fetch from the homeserver room directory

Fix https://github.com/matrix-org/matrix-public-archive/issues/80

```
RethrownError: Unable to fetch rooms from room directory (homeserver=http://localhost:8008/)
    searchTerm=, paginationToken=undefined, limit=9
    at matrix-public-archive\server\routes\room-directory-routes.js:55:13
    --- Original Error ---
    Error: HTTP Error Response: 500 Internal Server Error: {"errcode":"M_UNKNOWN","error":"Internal server error"}
        URL=http://localhost:8008/_matrix/client/v3/publicRooms?
        at checkResponseStatus (matrix-public-archive\server\lib\fetch-endpoint.js:21:11)
        at processTicksAndRejections (node:internal/process/task_queues:96:5)
        at async fetchEndpoint (matrix-public-archive\server\lib\fetch-endpoint.js:38:3)
        at async fetchEndpointAsJson (matrix-public-archive\server\lib\fetch-endpoint.js:63:15)
        at async fetchPublicRooms (matrix-public-archive\server\lib\matrix-utils\fetch-public-rooms.js:26:26)
        at async matrix-public-archive\server\tracing\trace-utilities.js:31:24
        at async matrix-public-archive\server\routes\room-directory-routes.js:45:62
```

### DIFF
--- a/server/routes/room-directory-routes.js
+++ b/server/routes/room-directory-routes.js
@@ -6,6 +6,7 @@ const urlJoin = require('url-join');
 const express = require('express');
 const asyncHandler = require('../lib/express-async-handler');
 
+const RethrownError = require('../lib/rethrown-error');
 const fetchPublicRooms = require('../lib/matrix-utils/fetch-public-rooms');
 const renderHydrogenVmRenderScriptToPageHtml = require('../hydrogen-render/render-hydrogen-vm-render-script-to-page-html');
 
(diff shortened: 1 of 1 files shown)


server/routes/room-directory-routes.js
```diff
@@ -31,19 +32,31 @@ router.get(
...
+          searchTerm,
+          paginationToken,
+          limit,
+        }
+      ));
+    } catch (err) {
+      throw new RethrownError(
+        `Unable to fetch rooms from room directory (homeserver=${matrixServerUrl})\n    searchTerm=${searchTerm}, paginationToken=${paginationToken}, limit=${limit}`,
+        err
+      );
```
MadLittleMods, line 58: This will still end up as a 500 error to the client but at least it's explained better.

I can't think of a better HTTP status code to return in this situation though 🤔

We could return a room directory with an empty list of rooms (hopefully with the error explained).

MadLittleMods: We can iterate on this later ⏩

PR #85, opened 2022-10-18
Have full room description readable from tooltip

MadLittleMods:
The `RoomCard`  clips text two lines before ellipsing. This will make the rest of the description readable if you hover over the description.

### DIFF
--- a/shared/views/RoomCardView.js
+++ b/shared/views/RoomCardView.js
@@ -70,7 +70,7 @@ class RoomCardView extends TemplateView {
           },
           [aliasOrRoomId]
         ),
-        t.p({ className: 'RoomCardView_topic' }, [vm.topic || '']),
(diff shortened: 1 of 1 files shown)


shared/views/RoomCardView.js
```diff
@@ -70,7 +70,7 @@ class RoomCardView extends TemplateView {
           },
           [aliasOrRoomId]
         ),
-        t.p({ className: 'RoomCardView_topic' }, [vm.topic || '']),
+        t.p({ className: 'RoomCardView_topic', title: vm.topic || null }, [vm.topic || '']),
```
MadLittleMods, line 73: Verified that `null` will cause no `title` attribute to be added to the element.

`undefined` will show up as `title="undefined"`

Issue #77, opened 2022-10-13
Off-by-one in calendar dates


MadLittleMods:
As discovered by @ara4n,

Starts on the right day but missing last day of the month (August should have 31 days)

<img src="https://user-images.githubusercontent.com/558581/195662082-bb71840f-5e22-4fee-99dd-f4790085d77f.png" width="50%">

MadLittleMods:
Interestingly, it has the proper amount of days for me:

![](https://user-images.githubusercontent.com/558581/195663479-7da36a8b-39a7-4f06-b2ac-8cf8fa90b99a.png)

MadLittleMods:
Reproducible with a London timezone which can be emulated with devtools -> Open the console drawer (press escape) -> Sensors (can be shown via the 3 vertical dot menu) -> Location

![](https://user-images.githubusercontent.com/558581/196553984-8f0e09fe-c557-458b-bc8a-1212e2da8dca.png)


![](https://user-images.githubusercontent.com/558581/196554205-e0f0a34f-a447-48ee-a8d4-0f248883e870.png)

MadLittleMods:
Fixed via https://github.com/matrix-org/matrix-public-archive/pull/86

This was a problem for any time zone greater than UTC+0

PR #86, opened 2022-10-18
Fix off-by-one calendar months with time zones that are greater than UTC+0

MadLittleMods:
Fix off-by-one calendar months with time zones that are greater than UTC+0 (GMT+0).

Fix https://github.com/matrix-org/matrix-public-archive/issues/77

Previously, we would calculate `lastDayOfTheMonthDate` in the local timezone because we were using the vanilla `new Date(year, month, 0)` constructor. For any timezone greater than `UTC+0` (like London UTC+1 or Korea UTC +9), this means that the date is a day-behind when we go back to UTC+0.

**Before:**

```
inputDate Fri, 29 Jul 2022 00:00:00 GMT 1659052800000
lastDayOfTheMonthDate Sat, 30 Jul 2022 23:00:00 GMT 1659222000000
lastDayOfTheMonth 30
```

**After**

```
inputDate Fri, 29 Jul 2022 00:00:00 GMT 1659052800000
lastDayOfTheMonthDate Sun, 31 Jul 2022 00:00:00 GMT 1659225600000
lastDayOfTheMonth 31
```

### Dev notes

It would be nice to be able to detect `Date` mis uses. We need to be careful to always use the UTC version of all of the date functions.

Can we carefully lint away usages of certain `Date` functions? We only want to allow `new Date(value)` with an integer (kinda seems like we need TypeScript for that) and related UTC member functions.

We definitely want to dis-allow [`new Date(year, monthIndex, day)` type constructors](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date) and any string parsing like `Date.parse(...)` and `new Date(dateString)` where the timezone can be assumed if it's not specified.

Random cool thought: Use [`semgrep`](https://github.com/returntocorp/semgrep) to find `Date` usages where we don't first use UTC stuff, etc

### DIFF
--- a/shared/views/CalendarView.js
+++ b/shared/views/CalendarView.js
@@ -21,9 +21,9 @@ function numDaysInMonthForDate(inputDate) {
   const lastDayOfTheMonthDate = new Date(
     Date.UTC(inputDate.getUTCFullYear(), inputDate.getUTCMonth() + 1, 0)
   );
-  const lastDayOfTheMonth = lastDayOfTheMonthDate.getUTCDate();
-  // The last day in the month is a proxy for how many days there are in that month
-  return lastDayOfTheMonth;
+  const lastDayNumberOfTheMonth = lastDayOfTheMonthDate.getUTCDate();
+  // The last day number in the month is a proxy for how many days there are in that month
+  return lastDayNumberOfTheMonth;
 }
 
 // Map from day of week to the localized name of the day
--- a/shared/views/CalendarView.js
+++ b/shared/views/CalendarView.js
@@ -18,10 +18,20 @@ function sameDay(date1, date2) {
 //
 // via https://stackoverflow.com/a/1184359/796832
 function numDaysInMonthForDate(inputDate) {
+  console.log('------------------------------------');
+  console.log('inputDate', inputDate.toUTCString(), new Date(inputDate).getTime());
   const lastDayOfTheMonthDate = new Date(
-    Date.UTC(inputDate.getUTCFullYear(), inputDate.getUTCMonth() + 1, 0)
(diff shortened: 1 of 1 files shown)


shared/views/CalendarView.js
```diff
@@ -147,9 +152,10 @@ class CalendarView extends TemplateView {
               }),
               (() => {
                 const todayTs = Date.now();
+                const numberOfDaysInMonth = numDaysInMonthForDate(calendarDate);
 
                 let dayNodes = [];
-                for (let i = 0; i < numDaysInMonthForDate(calendarDate); i++) {
+                for (let i = 0; i < numberOfDaysInMonth; i++) {
```
MadLittleMods, line 158: This avoids making the calculation for every iteration of the loop

shared/views/CalendarView.js
```diff
@@ -12,13 +12,18 @@ function sameDay(date1, date2) {
...
+// Get the number of days in the given month where the `inputDate` lies.
 //
 // via https://stackoverflow.com/a/1184359/796832
-function numDaysInMonthForDate(date) {
-  return new Date(date.getUTCFullYear(), date.getUTCMonth() + 1, 0).getUTCDate();
+function numDaysInMonthForDate(inputDate) {
+  // Month in JavaScript is 0-indexed (January is 0, February is 1, etc),
+  // but by using 0 as the day it will give us the last day of the prior
+  const lastDayOfTheMonthDate = new Date(
+    Date.UTC(inputDate.getUTCFullYear(), inputDate.getUTCMonth() + 1, 0)
```
MadLittleMods, line 22: The fix is here. We should only be manually constructing dates from a UTC helper first.

PR #88, opened 2022-10-19
Implement `ModalView` with native `<dialog>` element

MadLittleMods:
Implement `ModalView` with native `<dialog>` element. The main benefit we're getting from this refactor is semantics, `Escape` to close modal, and focus lock.

We have one existing modal with the Developer Options.

This refactor is happening so it can be used to add a new homeserver in the homeserver selector on the room directory page, https://github.com/matrix-org/matrix-public-archive/pull/87

### Todo

 - [x] Make sure the URL changes when you close the modal
    - [x] Closed via backdrop click
    - [x] Closed via `x` dismiss icon
    - [x] Closed via `Escape` keyboard shortcut
 - [x] Make sure the URL changes when you open the modal
 - [x] Make sure the modal opens when you page-load with `#/developer-options`

### Dev notes

 - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog
 - https://www.stefanjudis.com/blog/a-look-at-the-dialog-elements-super-powers/
 - https://twitter.com/argyleink/status/1511761383024521218
 - https://css-tricks.com/some-hands-on-with-the-html-dialog-element/


public/css/styles.css
```diff
@@ -279,38 +279,38 @@ body {
...
-
-.DeveloperOptionsView_modal {
-  z-index: 1;
-  overflow-y: auto;
+.ModalView_modal {
   width: 100%;
   max-width: 500px;
   max-height: 80%;
-  margin-left: 10px;
-  margin-right: 10px;
```
MadLittleMods, line 299: If we leave the `margin-left`/`margin-right` set, it makes the `dialog` not positioned in the middle of the screen.

#### Actual:



#### Expected:




But these were useful to make the modal not butt up against the very edge of the screen. It's not the end of the world though 🤔

MadLittleMods: It looks like we can get the same effect by using `max-width: min(500px, 100% - 20px);` ✅

Issue #92, opened 2022-10-20
Homeserver mismatch between homeserver selector and selected homeserver indicated by the `?homeserver` query parameter


MadLittleMods:
Visiting `/?search=&homeserver=foo.bar` will still leave you with the default configured `matrixServerName` selected in the homeserver selector UI on the room directory landing page.

There is mismatch between the homeserver selector UI and selected homeserver indicated by the `?homeserver` query parameter

**Example:** Visiting `/?search=&homeserver=foo.bar`, it should say `foo.bar` (not `gitter.im`)
![](https://user-images.githubusercontent.com/558581/196881536-c0128e8f-db3b-4e9b-a22d-12e275fcacfb.png)

PR #98, opened 2022-10-21
Page-load with the correct homeserver selected

MadLittleMods:
Page-load with the correct homeserver selected (according to `?homeserver`).

Fix https://github.com/matrix-org/matrix-public-archive/issues/92

Also makes sure that the `?homeserver` is always available somewhere in the list; whether that be in the available homeserver list or the added homeserver list depending on it someone cleared it out or never had it because they visited from someone else's link.

### Todo

 - [x] Fix problem where the "Add server" modal pops up again after adding a server and the page loads again
 - [x] Add test that page-loads with the correct option selected
 - [x] Make release of latest [Hydrogen scratch changes](https://github.com/vector-im/hydrogen-web/pull/653)

### DIFF
--- a/shared/lib/archive-history.js
+++ b/shared/lib/archive-history.js
@@ -28,22 +28,22 @@ class ArchiveHistory extends History {
     // downstream call of `urlRouter.attach()` which we do when bootstraping
     // everything.
     if (window.history) {
-      let replacingUrl = url;
+      let replacingUrl = window.location.search + url;
       // This is a way to make sure the hash gets cleared out
       if (url === '') {
-        replacingUrl = window.location.pathname;
+        replacingUrl = window.location.pathname + window.location.search;
       }
-      super.replaceUrlSilently(replacingUrl + window.location.search);
+      super.replaceUrlSilently(replacingUrl);
     }
   }
 
   pushUrlSilently(url) {
-    let replacingUrl = url;
+    let replacingUrl = window.location.search + url;
     // This is a way to make sure the hash gets cleared out
     if (url === '') {
-      replacingUrl = window.location.pathname;
+      replacingUrl = window.location.pathname + window.location.search;
     }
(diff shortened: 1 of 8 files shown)


shared/lib/archive-history.js
```diff
@@ -28,22 +28,22 @@ class ArchiveHistory extends History {
...
   pushUrlSilently(url) {
-    let replacingUrl = url;
+    let replacingUrl = window.location.search + url;
     // This is a way to make sure the hash gets cleared out
     if (url === '') {
-      replacingUrl = window.location.pathname;
+      replacingUrl = window.location.pathname + window.location.search;
     }
-    super.pushUrlSilently(replacingUrl + window.location.search);
+    super.pushUrlSilently(replacingUrl);
```
MadLittleMods, line 46: These changes should be split out. This fixes the problem of the hash and query parameter ordering and repeating.

shared/views/RoomDirectoryView.js
```diff
@@ -51,13 +47,37 @@ class RoomDirectoryView extends TemplateView {
...
+                );
+              }
+            );
+
+            let availableHomeserversOptGroup = text('');
+            if (availableHomeserverOptionElements.length > 0) {
+              availableHomeserversOptGroup = t.optgroup(
+                { label: 'Defaults' },
+                availableHomeserverOptionElements
+              );
```
MadLittleMods, line 67: Have to put this under a "Defaults" `<optgroup>` in order to be compatible with Hydrogen TemplateView's requiring a single root element.

I would ideally just return the array of `<option>` elements as before but we need the data binding to update the list when it changes. I don't want to use a `ListView` because it's cumbersome with it's observable list requirement.

shared/lib/archive-history.js
MadLittleMods: Split out to https://github.com/matrix-org/matrix-public-archive/pull/99

PR #102, opened 2022-10-21
Stop using `selectEl.value` as crutch over swapping `selected` attribute

MadLittleMods:
Stop using `selectEl.value` as crutch over swapping `selected` attribute

Related context: https://github.com/WebReflection/linkedom/issues/171#issuecomment-1287334938

Originally introduced in:

 - https://github.com/matrix-org/matrix-public-archive/pull/87
 - https://github.com/matrix-org/matrix-public-archive/pull/98

### DIFF
--- a/shared/views/RoomDirectoryView.js
+++ b/shared/views/RoomDirectoryView.js
@@ -54,7 +54,10 @@ class RoomDirectoryView extends TemplateView {
             const availableHomeserverOptionElements = vm.availableHomeserverList.map(
               (homeserverName) => {
                 return t.option(
-                  { value: homeserverName, selected: vm.homeserverSelection === homeserverName },
+                  {
+                    value: homeserverName,
+                    selected: (vm) => vm.homeserverSelection === homeserverName,
+                  },
                   homeserverName
                 );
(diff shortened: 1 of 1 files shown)


shared/views/RoomDirectoryView.js
```diff
@@ -169,8 +175,6 @@ class RoomDirectoryView extends TemplateView {
     t.mapSideEffect(
       (vm) => vm.homeserverSelection,
       (homeserverSelection, oldHomeserverSelection) => {
-        homeserverSelectElement.value = homeserverSelection;
```
MadLittleMods, line 172: Removing this introduces a bug where the `<select>` isn't updated by the time we submit the form after selection just below.

We have a similar problem with the hash as well and we just set it directly before submitting. Which is the same concept as what this did.

The alternative is to do a `setTimeout(() => {...}, 0)` and hope that everything is up to date by the next loop.

MadLittleMods: Fixed in https://github.com/matrix-org/matrix-public-archive/pull/105

Issue #91, opened 2022-10-20
Add `?via` query parameter with the current homeserver selected for rooms on the room directory landing page


MadLittleMods:
Add `?via` query parameter with the current homeserver selected for rooms on the room directory landing page.

It might be our first time for our archiver user to join the room when someone presses **View** from the room directory landing page.

PR #104, opened 2022-10-21
Link rooms with `?via` query parameter from room directory landing page

MadLittleMods:
Link rooms with `?via` query parameter from room directory landing page

Fix https://github.com/matrix-org/matrix-public-archive/issues/91

So we're always able to join the room from our origin homeserver to the remote room we don't know about yet.

### DIFF
--- a/shared/viewmodels/RoomDirectoryViewModel.js
+++ b/shared/viewmodels/RoomDirectoryViewModel.js
@@ -34,20 +34,6 @@ class RoomDirectoryViewModel extends ViewModel {
     this._homeserverUrl = homeserverUrl;
     this._homeserverName = homeserverName;
     this._matrixPublicArchiveURLCreator = matrixPublicArchiveURLCreator;
-    this._rooms = new ObservableArray(
-      rooms.map((room) => {
-        return {
-          roomId: room.room_id,
-          canonicalAlias: room.canonical_alias,
-          name: room.name,
-          mxcAvatarUrl: room.avatar_url,
-          homeserverUrlToPullMediaFrom: homeserverUrl,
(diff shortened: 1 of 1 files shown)


shared/viewmodels/RoomDirectoryViewModel.js
```diff
@@ -78,6 +64,23 @@ class RoomDirectoryViewModel extends ViewModel {
...
+        return {
+          roomId: room.room_id,
+          canonicalAlias: room.canonical_alias,
+          name: room.name,
+          mxcAvatarUrl: room.avatar_url,
+          homeserverUrlToPullMediaFrom: homeserverUrl,
+          numJoinedMembers: room.num_joined_members,
+          topic: room.topic,
+          archiveRoomUrl: matrixPublicArchiveURLCreator.archiveUrlForRoom(room.room_id, {
+            viaServers: [this.pageSearchParameters.homeserver],
```
MadLittleMods, line 78: Adding the `?via` query parameter messes with the client-side [`ArchiveHistory`](https://github.com/matrix-org/matrix-public-archive/blob/d19c08397a9c4b9a9be091b714a8a664ba66b88f/shared/lib/archive-history.js) and causes the URL to be duplicated.

`http://localhost:3050/!xxx:server/date/2022/09/20?via=my.synapse.server` -> `http://localhost:3050/!xxx:server/date/2022/09/20?via=my.synapse.serverhttp://localhost:3050/!xxx:server/date/2022/07/29`

MadLittleMods: Fixed in https://github.com/matrix-org/matrix-public-archive/pull/110

Issue #95, opened 2022-10-20
Fix room image avatars not showing on `RoomCard`s


MadLittleMods:
Fix room image avatars not showing on `RoomCard`s

They are currently just using the solid color fallback avatars with a letter.

![](https://user-images.githubusercontent.com/558581/197075354-6bc72202-1e5e-4ac4-b459-363c65c83d6e.png)

PR #133, opened 2022-11-10
Fix room image avatars not showing in the room directory

MadLittleMods:
Fix room image avatars not showing in the room directory

Fix https://github.com/matrix-org/matrix-public-archive/issues/95

Before | After
--- | ---
. | .


shared/views/RoomCardView.js
```diff
@@ -7,13 +7,13 @@ class RoomCardView extends TemplateView {
   render(t, vm) {
     const avatarViewModel = new AvatarViewModel({
       homeserverUrlToPullMediaFrom: vm.homeserverUrlToPullMediaFrom,
-      avatarUrl: vm.avatarUrl,
+      avatarUrl: vm.mxcAvatarUrl,
```
MadLittleMods, line 10: Main fix is here

Issue #7, opened 2022-02-24
Add hour range time picking UI


MadLittleMods:
Add hour range time picking UI

Related to https://github.com/matrix-org/matrix-public-archive/issues/3

Probably just something simple like the Gitter UI:

`/<community>/<room>/archives/2021/02/09/0-1`
![](https://user-images.githubusercontent.com/558581/155483515-bf44d7fe-0f5a-405d-9dd3-61a7206ed752.png)

### Problems

How can we deal with a massive amount of messages? An hour chunk could easily contain 100, 500 messages which reaches our limit. Do we keep scaling the time increment down until we fit in the message limit?

How do we scale the UI when there are 500 messages per each second? How does the URL scale/represented at those small time increments? Perhaps we give up at a certain increment.

Perhaps a nice sliding window to scrub along the timeline of a day. The window size is based on what we could render on the current page which can be moved anywhere else in the day. If the density of messages changes on the next page, then the window size will adjust for them to continue. We can always truncate the window short to what fits on the page.

<img width="674" alt="" src="https://user-images.githubusercontent.com/558581/198814607-f6c1e84c-bd6d-4445-b482-3b7c3158d51a.png">

### Potential solution

URL structure: `/r/test-room1:my.synapse.server/date/2022/11/11T12:05`
 - Similar to `new Date().toISOString()` -> `2022-11-14T22:41:14.721Z`

This will show messages from that time backwards until we reach the message limit.

We can show UI like this to pick a time in UTC and have them easily compare to their local timezone.

https://dribbble.com/shots/14590546-Time-Zone-Translate
<img src="https://user-images.githubusercontent.com/558581/201458106-b0e03a37-2af9-499c-b88b-bed1315838ec.png" width="400">

### Design exploration

 - Timeline scrubbing
 - Time range selection
 - Time series navigator
 - Time slice
 - Sliding window of time
 - Time drilldown

---

Jaeger tracing UI:
<img width="674" alt="" src="https://user-images.githubusercontent.com/558581/198814607-f6c1e84c-bd6d-4445-b482-3b7c3158d51a.png">

https://dribbble.com/shots/5502990-Time-Scrubber
![](https://user-images.githubusercontent.com/558581/201458068-77552278-cdc0-4fb5-8692-ac4054f3b50e.gif)

https://dribbble.com/shots/14590546-Time-Zone-Translate
[...]

PR #139, opened 2022-11-15
Add a way to select time of day

MadLittleMods:
Add a way to select time of day. This includes the nice time picker and all of the backend necessary to handle the time slicing and precision.

 - Fix https://github.com/matrix-org/matrix-public-archive/issues/7
 - A URL with time looks like
    - `/r/too-many-messages-on-day:my.synapse.server/date/2022/11/16T23:59`
    - Or when more precision is required (seconds): `/r/too-many-messages-on-day:my.synapse.server/date/2022/11/16T23:59:59`
 - Add new custom time picker/scrubber (pictured below) with momentum scrubbing
    - Native built-in `<input type="time">` for easier picking if you prefer that and accessibility.
    - Uses localized time strings
    - Design inspired by Thiago Sanchez's *Time Zone Translate* concept, https://dribbble.com/shots/14590546-Time-Zone-Translate

Native built-in `<input type="time">` for easier picking if you prefer that and accessibility:

Uses localized time strings (Firefox pictured):

Time selector with localized strings in Firefox

It seems Chrome has a bug where it doesn't localize the AM/PC part of the `<input type="time">` but that's a problem with Chrome itself as it works just fine with Firefox above

Time selector with localized strings in Chrome where it literally says "AM/PM" instead of the localized version that we see printed when using `Date.toLocaleTimeString(...)`

### Dev notes

Test rooms:

 - http://192.168.1.149:3050/r/too-many-messages-on-day:my.synapse.server/date/2022/11/16
 - http://192.168.1.149:3050/r/test-room1:my.synapse.server/date/2022/11/15

---

 - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString
 
 Ruler ticks CSS:
 
  - https://codepen.io/thebabydino/pen/RwjWrKz
  - https://codepen.io/webandapp/details/MJqMyE
  
Scroll snap and drag to scroll:

 - https://web.dev/css-scroll-snap/
 - https://dev.to/wolffe/i-created-a-draggable-carousel-with-momentum-scrolling-and-mobile-support-using-vanilla-javascript-17he
    - https://getbutterfly.com/how-to-create-a-draggable-carousel-using-vanilla-javascript/
 - https://ariya.io/2011/10/flick-list-with-its-momentum-scrolling-and-deceleration
 - https://swiperjs.com/ 

 Random:
 
 - http://forbeslindesay.github.io/express-route-tester/
    - Express uses [`path-to-regexp`](https://www.npmjs.com/package/path-to-regexp)
 - https://alvaromontoro.com/blog/67970/drawing-a-triangle-with-css
 - Scroll without scrollbars, https://stackoverflow.com/a/38994837/796832

### Todo

Frontend:

 - [x] Change time as you scroll around the timeline
 - [x] In the `TimeSelector`, change time text as you manipulate the `TimeSelector`
 - [x] In the `TimeSelector`, using the `<input type="time">` should also change the time
 - [x] In the `TimeSelector`, make sure after using the `<input type="time">` to select a time, it continues to update if you scrub around more
 - [x] In the `TimeSelector`, show magnifier bubble around range of messages in the timeline on this page
 - [x] In the `TimeSelector`, show "Go" button ~~when outside of that visible range (meaning they picked a spot that is somewhere new to visit and see)~~ (we can just simply show the "Go" button always)
 - [ ] ~~In the `TimeSelector`, add some snapping to the far end of the current range so people can easily jump to the next chunk seamlessly~~ (not necessary for this iteration)
 - [x] Only show `TimeSelector` if visiting page that is showing messages within a day (`Txxx` in URL)
 - [ ] ~~Prevent extra precision if it's not needed (ensure this works as you manipulate the `TimeSelector`)~~ We will just assume whatever precision from the initial date in the URL
 - [ ] ~~Jump links use precision~~ (no longer necessary)
 - [x] When using mobile device emulation, the scrubber scroll position isn't initialized on page load
    - When the right-panel is collapsed and the page is loaded, it doesn't work because the time selector is hidden and has `scrollWidth`/`clientWidth` as `0` which evaluates the `scrollLeft` to `0`.
    - When the right-panel is expanded and the page is loaded, it doesn't work because the `currentTopPositionEventEntry` is never set because the timeline isn't shown and the `IntersectionObserver` never triggers so `activeDate` never updates and the side-effect to update the scrubber `scrollLeft` position doesn't run. And the side-effect doesn't update the `scrollLeft` the first time it runs because it runs before the view is mounted to the DOM.
 - [ ] ~~Stretch: Add `+1` and `-1` day indicators in the local time. Someone's timezone could cause them to be a whole day ahead or behind.~~
    - -> Tracked separately as https://github.com/matrix-org/matrix-public-archive/issues/163

Backend:

 - [ ] ~~Redirect to specific time if too many messages on that day~~ (doesn't make any sense)

Tests:

 - [x] Add test for loading a day with too many messages for the day
 - [x] Add test for loading a day with too many messages for the hour
 - [x] Add test for loading a day with too many messages for the minute
 - [x] Add test for loading a day with too many messages for the second (gives up)
 - [x] Add test to jump forwards to the latest messages in the room, ensure you can see the latest message
 - [x] Add test for jumping forward into too many messages
 - [x] Add test for jumping backward into too many messages
 - [x] Add test for jumping forward from one day with too many messages into the next day too many messages
 - [x] Add test for jumping backward from one day with too many messages into the previous day too many messages
 - [x] Add test for jumping forward from too many messages into more too many messages on the same day
 - [x] Add test for jumping backward from too many messages into more too many messages on the same day
 - [ ] ~~Add test for jumping forward from one day too many messages into the next with a normal amount of messages~~ (turns out to be similar to others)
 - [ ] ~~Add test for jumping backward from one day too many messages into the previous day with a normal amount of messages~~ (turns out to be similar to others)
 - [ ] ~~Add test for jumping forward from too many messages into normal amount of messages on the same day (does this make sense? maybe too many in hour to fitting in an hour)~~ (I think this is fairly covered by our jump precision tests)
 - [ ] ~~Add test for jumping backward from too many messages into normal amount of messages on the same day (does this make sense? maybe too many in hour to fitting in an hour)~~ (I think this is fairly covered by our jump precision tests)
 - [x] Add test to make sure jumping forward over many days is as expected
 - [x] Add test to make sure jumping backward with many small days is as expected
 - [ ] ~~Add test for no more messages forward (shows empty view)~~ (this one isn't possible to come across as a room will always have a create event for example which we will see for any given day backwards.)
 - [x] Add test for no more messages backward (shows empty view)
 - [x] Add test for visiting a room before the room was created, will show empty view so the user can paginate forward
 - [x] Add test for showing time selector UI when showing only showing events from same day
 - [x] Add test for showing time selector UI when showing only showing events from same day but paginated forward into days without any messages
 - [x] Add test for to make sure the time selector UI is hidden when showing events from multiple days

---

Tests:

 - [ ] ~~Make sure jump to previous link timestamp does not come after the last message timestamp~~
    - Imagine someone imported some messages out of order with the wrong timestamp. We don't want that to throw someone into a giant unexpected jump. Try to maintain the course through the days
    - This can happen in our tests for example because we're using timestamp massaging to get consistent dates but the room creation events are still from the current day.
    - -> Seems niche and we can deal with this a later point
 - [ ] ~~Make sure jump to next link timestamp does not come before the first message timestamp~~
    - -> Seems niche and we can deal with this a later point


test/e2e-tests.js
```diff
@@ -656,235 +784,695 @@ describe('matrix-public-archive', () => {
...
+            // because there are more than 5 messages in between day4 and day2.
+            //
+            // Even though there is overlap between the pages, our scroll continues from
+            // the event where the 1st page starts.
+            //
+            // 1 <-- 2 <-- 3 <-- 4 <-- 5 <-- 6 <-- 7 <-- 8 <-- 9 <-- 10 <-- 11 <-- 12
+            // [day1       ]     [day2       ]     [day3       ]     [day4          ]
+            //       [1st page               ]
+            //                               |--jump-fwd-4-messages-->|
+            //                         [2nd page               ]
```
MadLittleMods, line 803: Maybe we can refactor this whole test format into something more compact and readable

Need to think about how to express the action(prev/next), expected precision

```
{
  events: [[1, 2], [3, 4, 5, 6, 7, 8], [9, 10, 11, 12, 13, 14]],
   page1:                              [9, 10, 11, 12, 13, 14],
   page2:                [5, 6, 7, 8 ,  9, 10],
}
```

server/routes/room-routes.js
```diff
@@ -153,33 +199,134 @@ router.get(
...
+        const fromSameSecond = areTimestampsFromSameSecond(currentRangeEndTs, tsForClosestEvent);
+
+        console.log('fromSameDay', fromSameDay);
+        console.log('fromSameHour', fromSameHour);
+        console.log('fromSameMinute', fromSameMinute);
+        console.log('fromSameSecond', fromSameSecond);
+
+        // The closest event is from the same second we tried to jump from. Since we
+        // can't represent something smaller than a second in the URL (we could do MS
+        // but it's a concious choice to make the URL cleaner), we will need to just
```
MadLittleMods, line 284: I think we could let the URL support `ms` as long as we got rid of the `/date/2022/11/16T23:59:59` default. That should be equivalent to `/date/2022/11/16` anyway

MadLittleMods: > We can also assume that `T23:59:59` is equivalent to `T23:59:59.999` (instead of `T23:59:59.000`)

This is flawed and probably not good to do because then how do we represent the `T23:59:59.000` case.

Regardless, we should be able to coalesce to `/date/2022/11/16` somehow

test/e2e-tests.js
MadLittleMods: Played with this a bit during a refactor and didn't end up with anything better that is compact like this.

This kind of thing would be best edited with some simple custom UI but that's out of scope. It would be nice to make interactable visualizations inline to the code here.

server/routes/room-routes.js
MadLittleMods: 👍 Removed the `T23:59:59.000` stuff

shared/lib/timestamp-utilities.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
+
+const assert = require('matrix-public-archive-shared/lib/assert');
+const { MS_LOOKUP } = require('matrix-public-archive-shared/lib/reference-values');
+const { ONE_DAY_IN_MS, ONE_HOUR_IN_MS, ONE_MINUTE_IN_MS, ONE_SECOND_IN_MS } = MS_LOOKUP;
+
+// TODO: It would be good to add a few test for these
```
MadLittleMods, line 7: Add some tests for these

MadLittleMods: Added ✅ -> https://github.com/matrix-org/matrix-public-archive/pull/139/commits/2db3c4a93bb3ba7877e3ba74b3ea34f55630428a

test/e2e-tests.js
MadLittleMods: This dream is coming true in https://github.com/matrix-org/matrix-public-archive/pull/164 🎉

Issue #141, opened 2022-11-17
Error with Node.js v19.1.0: `Cannot set property crypto of #<Object> which has only a getter` when opening http://localhost:3050/


sideshowbarker:
After install, running `npm run start-dev` and then opening http://localhost:3050/ in my browser, I get this error:

```
RethrownError: Failed to render Hydrogen to string. In order to reproduce, feed in these arguments into `renderHydrogenToString(...)`:
[...]

PR #143, opened 2022-11-18
Only assign vmContext.global.crypto if not already global

sideshowbarker:
Fixes https://github.com/matrix-org/matrix-public-archive/issues/141

Node.js v19 has `crypto` set on the global already, so this change causes `vmContext.global.crypto` to be assigned only if `vmContext.global.crypto` isn’t already defined.

Otherwise, without this change, the room directory fails to render if Node.js v19+, and instead “TypeError: Cannot set property crypto of `#<Object>` which has only a getter” gets thrown.


test/e2e-tests.js
```diff
@@ -889,6 +889,12 @@ describe('matrix-public-archive', () => {
       });
     });
 
+    describe('Room directory renders as expected', () => {
+      it('renders the room directory', async () => {
+        await getTestClientForHs(testMatrixServerUrl1);
+      });
+    });
```
MadLittleMods, line 896: By "and a quick test", I just meant make sure it runs for you in Node.js 19.

Sorry this wasn't clear 🙇. We don't need this test. We already have many tests for making sure the archive and room directory works. We just don't run the test suite against Node.js 19 to catch it.

And I don't think we need to update the test matrix to also run with Node.js 19 yet. We can fix up the little ad-hoc things for now.

server/hydrogen-render/render-hydrogen-to-string-unsafe.js
```diff
@@ -47,7 +47,11 @@ function createDomAndSetupVmContext() {
   vmContext.global.DOMParser = dom.DOMParser;
   // Make sure `webcrypto` exists since it was only introduced in Node.js v17
   assert(crypto.webcrypto);
-  vmContext.global.crypto = crypto.webcrypto;
+  // Only assign vmContext.global.crypto if it’s undefined
+  // (Node.js v19 has crypto set on the global already)
```
MadLittleMods, line 51: Great comment 👍

Just suggesting we change the magic apostrophe to a normal one
```suggestion
  // Only assign vmContext.global.crypto if it's undefined
  // (Node.js v19 has crypto set on the global already)
```

test/e2e-tests.js
sideshowbarker: > By ["and a quick test"](https://github.com/matrix-org/matrix-public-archive/issues/141#issuecomment-1319383479), I just meant make sure it runs for you with Node.js 19.

aha 😀

OK — as far as that goes, I can confirm that is runs as expected with Node.js v19.1.0

> Sorry this wasn't clear 🙇.

No worries — it gave me to a chance to get a test environment set up, and to look through the test-harness code, and to actually run the tests. So it was all win for me

sideshowbarker: …and just now pushed a change that removes the test

MadLittleMods:
Thanks for the contribution @sideshowbarker 🙂 🐆

🙌 for Node.js 19 support and for things working out of the box!

PR #147, opened 2022-11-18
Enable CORS support

sideshowbarker:
This change enables CORS support in the archive — to allow web developers to create web applications with frontend JavaScript code that can fetch pages from the archive (for example, for scraping content from chat logs).

Otherwise, without this change, web developers can’t create web apps with frontend JavaScript that can fetch chat logs from the archive and then consume the content of the logs.

It’s imaginable that web developers may find use cases for consuming the chat logs in the archive from frontend JavaScript code — at the simplest level, web apps that fetch and scrape logs to get data out of them or to pull out particular snippets from the logs.

Developers can anyway already scrape the contents of the archive — by using server-side programming languages or by using `curl` or whatever from the command line. They just can’t do the same from frontend JavaScript code, unless CORS support is enabled.

### DIFF
--- a/server/routes/install-routes.js
+++ b/server/routes/install-routes.js
@@ -2,6 +2,7 @@
 
 const path = require('path');
 const express = require('express');
+const cors = require('cors');
 const asyncHandler = require('../lib/express-async-handler');
 
 const { handleTracingMiddleware } = require('../tracing/tracing-middleware');
@@ -15,6 +16,7 @@ function installRoutes(app) {
   app.use(handleTracingMiddleware);
   app.use(preventClickjackingMiddleware);
   app.use(contentSecurityPolicyMiddleware);
+  app.use(cors());
 
   let healthCheckResponse;
   app.get(
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "@opentelemetry/resources": "^1.3.1",
         "@opentelemetry/sdk-trace-base": "^1.3.1",
         "@opentelemetry/semantic-conventions": "^1.3.1",
+        "cors": "^2.8.5",
         "dompurify": "^2.3.9",
         "express": "^4.17.2",
         "hydrogen-view-sdk": "npm:@mlm/hydrogen-view-sdk@^0.25.0-scratch",
@@ -2531,6 +2532,18 @@
         "url": "https://opencollective.com/core-js"
       }
     },
+    "node_modules/cors": {
+      "version": "2.8.5",
+      "resolved": "https://registry.npmjs.org/cors/-/cors-2.8.5.tgz",
+      "integrity": "sha512-KIHbLJqu73RGr/hnbrO9uBeixNGuvSQjul/jdFvS/KFSIH1hWVd1ng7zOHx+YrEfInLG7q4n6GHQ9cDtxv/P6g==",
+      "dependencies": {
+        "object-assign": "^4",
+        "vary": "^1"
+      },
+      "engines": {
+        "node": ">= 0.10"
+      }
+    },
     "node_modules/cross-spawn": {
       "version": "7.0.3",
       "dev": true,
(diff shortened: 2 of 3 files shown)


server/routes/install-routes.js
```diff
@@ -15,6 +16,7 @@ function installRoutes(app) {
   app.use(handleTracingMiddleware);
   app.use(preventClickjackingMiddleware);
   app.use(contentSecurityPolicyMiddleware);
+  app.use(cors());
```
MadLittleMods, line 19: I'm not sure we want to support this (genuine question) 🤔. I feel like I don't have enough data points and the right knowledge to really make the taste-making decision here either. You can already fetch messages in the room yourself via the Matrix API.

Since the app is stateless and you can't do any actions (we only fetch information), we aren't concerned with the usual [CSRF](https://owasp.org/www-community/attacks/csrf) attacks.

We already prevent people from embedding the archive in an iframe via [`server/routes/prevent-clickjacking-middleware.js`](https://github.com/matrix-org/matrix-public-archive/blob/6b493ff807767fb744b328c14abbe6dfc213dadf/server/routes/prevent-clickjacking-middleware.js).

It wouldn't be fun for attackers to use unsuspecting people just browsing another site to be used as DDOS clients against Matrix Public Archive which also means against the Matrix homeserver. The same could be said for guest access with the Matrix API but at least the homeserver could block the source directly instead of blocking the Matrix Public Archive if someone starts to abuse things. Of course the same attack could be done outside of the browser, but the browser is an easy target.

Do you actually have a use case for this yet? I am all for client-side webapps so it's not like I am totally against this but without something tangible, I'm inclined to defer the decision to later.

sideshowbarker: > Do you actually have a use case for this yet?

Yes — the specific use case I have in mind, to describe it briefly, is for generating meeting minutes; that is, for a web service (frontend web app) that generates meeting minutes excerpted from the Matrix public archive.

In more detail: We (W3C) are going to end up needing a replacement for our [RRSAgent](https://www.w3.org/2002/03/RRSAgent) IRC bot that all our working groups currently use for (1) logging their raw meeting minutes and then (2) generating formatted HTML minutes from the raw minutes. It’s very similar to the Fedora project’s [Zodbot](https://fedoraproject.org/wiki/Zodbot) IRC bot.

Once the Matrix public archive launches, we’d no longer need the logging function of our RRSAgent (<span>#</span>1 above) but will instead just need some way to generate the formatted HTML minutes (<span>#2</span> above).

So what we’re anticipating is that we could have frontend web app (web service) that generates meeting minutes by excerpting them from the Matrix public archive logs — based on it looking for the same annotations that we already use now for demarcating meetings with our RRSAgent tool.

If the public archive isn’t CORS-enabled, then we’d not be able to use a frontend web app to generate the meeting minutes, but would instead be limited to needing to do it with some additional backend server-side runtime.

> I am all for client-side webapps so it's not like I am totally against this but without something tangible, I'm inclined to defer the decision to later.

I’m 100% fine with deferring it — there’s certainly no rush. But that said, once the public archive actually launches, it could be provisionally CORS-enabled, and we could watch to what unanticipated problems it may cause, and then disable CORS if there’s anything found that seems to be causing problems.

> It wouldn't be fun for attackers to use unsuspecting people just browsing another site to be used as DDOS clients against Matrix Public Archive which also means against the Matrix homeserver. … Of course the same attack could be done outside of the browser, but the browser is an easy target.

Right — but unless I’m misunderstanding the attack you have mind, I don’t think keeping CORS disabled would prevent it. That is, since the CORS protocol doesn’t cause any _request_ blocking, then CORS-enabling the public archive wouldn’t cause the server to start getting requests it wouldn’t already be getting. It would instead just cause browsers to un-block cross-origin access to any _responses_ that browsers got back from the archive.

In other words, if some attacker wanted to use unsuspecting people to DDOS the public archive from frontend code, keeping CORS disabled would do nothing to prevent that, as far as I can see. Specifically, lack of CORS wouldn’t prevent the attacker from _sending_ as many requests from as many frontend clients as they wanted. It’d prevent the attacker’s frontend from being able to access any of the _responses_ — but if the attacker’s goal is just to DDOS the archive, they presumably wouldn’t care about the responses anyway.

MadLittleMods: > Right — but unless I’m misunderstanding the attack you have mind, I don’t think keeping CORS disabled would prevent it. That is, since the CORS protocol doesn’t cause any request blocking

Thanks for pushing back here and for explaining. You're right! For my own reference: CORS doesn't stop [*simple requests*](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests) from reaching the destination server since they don't have any preflight `OPTIONS` request beforehand.


> Yes — the specific use case I have in mind, to describe it briefly, is for generating meeting minutes; that is, for a web service (frontend web app) that generates meeting minutes excerpted from the Matrix public archive.

Your use case seems real enough to justify this 👍

MadLittleMods:
Thanks for the contribution @sideshowbarker 🐔

sideshowbarker:
> Thanks for the contribution @sideshowbarker 🐔

Cheers and thanks for doing all the actual work toward bringing the full-featured public archive to reality — really looking forward to having it

Issue #89, opened 2022-10-19
Add NSFW safe search filter to room directory landing page


MadLittleMods:
Add NSFW safe search filter to room directory landing page.

We can put placeholder `RoomCards` for any rooms caught in the filter. This way our grid of rooms is even, we don't throw off pagination by trying to fetch more than needed and filter down, people are aware of content they are not seeing, and we don't upset our Google SEO rankings as much.

![](https://user-images.githubusercontent.com/558581/196606265-51f4b87e-6769-4475-8b1d-2ac7525eac19.png)

The Safe search toggle control above the grid of rooms:

![](https://user-images.githubusercontent.com/558581/202874262-372acf8d-14bd-46f0-a086-1d54c6cd1852.png)

### What is considered NSFW?

Any room with `NSFW`, `porn`, `nudes` (maybe other keywords, `sex`, `18+`) in the title or topic/description. This won't be exhaustive or completely effective but seems like it would do a decent job.

### Dev notes

Default to safe search on which keeps it out of the query parameters by default.

---

Checkbox styles look something like http://jsfiddle.net/MadLittleMods/x049hLb7/

#### Prior art

Good language and semantics around how Google explains safe search.

![](https://user-images.githubusercontent.com/558581/196607888-0453b60e-8c29-4f9f-b03c-7d027839e324.png)

Jem256:
hey @MadLittleMods I'd like to give this a try. But I'd like some clarification on what the t and vm are in the code `render(t, vm)` and how the app is doing a search for the rooms

MadLittleMods:
> clarification on what the t and vm are in the code `render(t, vm)`

@Jem256 It's from Hydrogen so you'll get a better picture by exploring that codebase first, see [`vector-im/hydrogen-web` -> `doc/UI/ui.md`](https://github.com/vector-im/hydrogen-web/blob/dbbbb1c29ababbcbe3f933b2a9424e3c0fc00ec5/doc/UI/ui.md):

> In views based on `TemplateView`, you will see a render method with a `t` argument.  
> `t` is `TemplateBuilder` object passed to the render function in `TemplateView`. It also takes a data object to render and bind to, often called `vm`, short for view model from the MVVM pattern Hydrogen uses.
>
[...]

PR #155, opened 2022-12-28
Safe search filter

Jem256:
Fix: #89 

I would like a review of my work, whether I'm going in the right direction as well as help with how I can get the icon action to dispatch the filter function I have written

### DIFF
--- a/public/css/room-directory.css
+++ b/public/css/room-directory.css
@@ -122,7 +122,7 @@
   color: #eeeeee;
 }
 
-.RoomDirectoryView_homeserverSelectSection {
+.RoomDirectoryView_safeSearchSection, .RoomDirectoryView_homeserverSelectSection {
   display: flex;
   align-items: center;
   margin-top: 16px;
@@ -262,7 +262,6 @@
 .RoomDirectoryView_codeBlock {
   overflow: auto;
   padding: 0.5em 0.8em;
-
   background-color: #f6f6f6;
   border: 1px solid rgb(233, 233, 233);
   border-radius: 4px;
@@ -289,7 +288,7 @@
   }
 }
 
-.RoomCardView {
+.FilterCardView, .RoomCardView {
   overflow: hidden;
   display: flex;
   flex-direction: column;
@@ -299,6 +298,10 @@
   list-style: none;
   border: 1px solid rgba(54, 139, 214, 30%);
   border-radius: 8px;
+}
+
+.FilterCardView {
+  background-color: #abb5be;
 }
 
 .RoomCardView_header {
@@ -350,6 +353,16 @@
   text-overflow: ellipsis;
 }
 
+.FilterCardView_topic {
+  display: -webkit-box;
+  height: 2.4em;
+  margin-top: 8px;
+  margin-bottom: 8px;
+
+  line-height: 1.2em;
+  -webkit-box-orient: vertical;
+}
+
 .RoomCardView_footer {
   display: flex;
   /**
@@ -365,6 +378,13 @@
 .RoomCardView_footerInner {
   display: flex;
   justify-content: space-between;
+  align-items: center;
+  width: 100%;
+}
+
+.FilterCardView_footerInner {
+  display: flex;
+  justify-content: flex-end;
   align-items: center;
   width: 100%;
 }
@@ -408,4 +428,25 @@
 .RoomCardView_viewButtonWrapperLink:focus > .RoomCardView_viewButton {
   background-color: #2774c2;
   color: #ffffff;
+}
+
+.RoomDirectoryView_safeSearchInput {
+  -moz-appearance: none;
+  -webkit-appearance: none;
+  -o-appearance: none;
+  appearance: none;
+  position: relative;
+  width: 40px;
+  height: 20px;
+  background-color: #ffffff;
+  border: 1px solid rgba(200, 200, 200, 1);
+  box-shadow: inset -20px 0px 0px 0px rgba(200, 200, 200, 1);
+  border-radius: 10px;
+  transition: all 0.2s ease;
+  right: 0;
+}
+
+.RoomDirectoryView_safeSearchInput:checked {
+  order: 1px solid rgba(80, 134, 240, 1);
+  box-shadow: inset 20px 0px 0px 0px rgba(80, 134, 240, 1);
 }
--- a/public/css/room-directory.css
+++ b/public/css/room-directory.css
@@ -431,19 +431,17 @@
 }
 
 .RoomDirectoryView_safeSearchInput {
-  -moz-appearance: none;
-  -webkit-appearance: none;
-  -o-appearance: none;
-  appearance: none;
   position: relative;
+  right: 0;
   width: 40px;
   height: 20px;
+  
+  appearance: none;
   background-color: #ffffff;
   border: 1px solid rgba(200, 200, 200, 1);
   box-shadow: inset -20px 0px 0px 0px rgba(200, 200, 200, 1);
   border-radius: 10px;
   transition: all 0.2s ease;
-  right: 0;
 }
 
 .RoomDirectoryView_safeSearchInput:checked {
--- a/public/css/room-directory.css
+++ b/public/css/room-directory.css
@@ -122,7 +122,8 @@
   color: #eeeeee;
 }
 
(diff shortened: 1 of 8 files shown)


public/css/room-directory.css
```diff
@@ -409,3 +429,24 @@
...
+  box-shadow: inset -20px 0px 0px 0px rgba(200, 200, 200, 1);
+  border-radius: 10px;
+  transition: all 0.2s ease;
+  right: 0;
+}
+
+.RoomDirectoryView_safeSearchInput:checked {
+  order: 1px solid rgba(80, 134, 240, 1);
+  box-shadow: inset 20px 0px 0px 0px rgba(80, 134, 240, 1);
+}
```
MadLittleMods, line 452: We need to run Prettier over the `.css` styles as well

CI failure: https://github.com/matrix-org/matrix-public-archive/actions/runs/3794941206/jobs/6453539585#step:5:10

public/css/room-directory.css
```diff
@@ -409,3 +429,24 @@
...
+  position: relative;
+  width: 40px;
+  height: 20px;
+  background-color: #ffffff;
+  border: 1px solid rgba(200, 200, 200, 1);
+  box-shadow: inset -20px 0px 0px 0px rgba(200, 200, 200, 1);
+  border-radius: 10px;
+  transition: all 0.2s ease;
+  right: 0;
+}
```
MadLittleMods, line 447: I haven't checked this out locally to see if these styles are even relevant but I would reformat what's here to:

```suggestion
.RoomDirectoryView_safeSearchInput {
  position: relative;
  right: 0;
  width: 40px;
  height: 20px;
  
  appearance: none;
  background-color: #ffffff;
  border: 1px solid rgba(200, 200, 200, 1);
  box-shadow: inset -20px 0px 0px 0px rgba(200, 200, 200, 1);
  border-radius: 10px;
  transition: all 0.2s ease;
}
```

shared/views/FilterCardView.js
```diff
@@ -0,0 +1,76 @@
+'use strict';
+
+const { TemplateView, AvatarView } = require('hydrogen-view-sdk');
+const AvatarViewModel = require('../viewmodels/AvatarViewModel');
+
+class FilterCardView extends TemplateView {
```
MadLittleMods, line 6: Would it be easier to combine this logic into `RoomCardView`?

I'm not totally opposed to having something separate but feels like could be combined

MadLittleMods, line 6: `FilterCardView` doesn't really describe what this single purpose component is for.

Name suggestion:

```suggestion
class NsfwCardView extends TemplateView {
```

shared/views/FilterCardView.js
```diff
@@ -0,0 +1,76 @@
...
+              {
+                className: 'RoomCardView_viewButtonWrapperLink',
+                href: vm.archiveRoomUrl,
+                title: `View the ${displayName} room`,
+              },
+              t.span(
+                {
+                  className: 'RoomCardView_viewButton',
+                },
+                ['View']
```
MadLittleMods, line 66: We need to disable the **View** link/button if safe search is on.

shared/views/RoomDirectoryView.js
```diff
@@ -17,13 +20,39 @@ class RoomDirectoryView extends TemplateView {
...
+              .toLowerCase()
+              .includes(
+                element.toLowerCase() || topic.toLowerCase().includes(element.toLowerCase())
+              )
+          ) {
+            return true;
+          }
+          return false;
+        });
+        if (contains === true) {
```
MadLittleMods, line 53:
```suggestion
        if (isRoomNsfw) {
```

shared/views/RoomDirectoryView.js
```diff
@@ -17,13 +20,39 @@ class RoomDirectoryView extends TemplateView {
...
+          return new FilterCardView(room);
```
MadLittleMods, line 54: Does the room-list properly update if you toggle safe-search on and off? My guess is no and it might be tricky to force Hydrogen to re-render here since we're using a `ListView`

shared/views/RoomDirectoryView.js
```diff
@@ -17,13 +20,39 @@ class RoomDirectoryView extends TemplateView {
...
+            name
+              .toLowerCase()
+              .includes(
+                element.toLowerCase() || topic.toLowerCase().includes(element.toLowerCase())
+              )
+          ) {
+            return true;
+          }
+          return false;
+        });
```
MadLittleMods, line 52: This currently matches words like `sextant` where the substring is in there but it's not what we're after.

I think we probably want to go with something like this which uses the `\b` regex word-boundary character

```suggestion
        const isRoomNsfw = NSFW_KEYWORDS.some((nsfwKeyword) => {
          const nsfwRegex = new RegExp(`/\\b${nsfwKeyword}\\b/i`);
          return room.name?.match(nsfwRegex) || room.topic?.match(nsfwRegex);
        });
```

shared/views/RoomDirectoryView.js
```diff
@@ -17,13 +20,39 @@ class RoomDirectoryView extends TemplateView {
...
     }
 
+    function toggleSafeSearchLabel() {
+      var text = document.getElementById('checkBox_label');
+      if (text.innerHTML === 'Safe Search is on') {
+        text.innerHTML = 'Safe Search is off';
+      } else {
+        text.innerHTML = 'Safe Search is on';
+      }
+    }
```
MadLittleMods, line 30: Instead of manipulating the DOM directly, we should update the view model and have Hydrogen re-render for us.

As an example see:

 - [`shared/views/DeveloperOptionsContentView.js#L26`](https://github.com/matrix-org/matrix-public-archive/blob/f95086df0d22393376916e087bc0ced09ba3f389/shared/views/DeveloperOptionsContentView.js#L26)
  - [`shared/viewmodels/DeveloperOptionsContentViewModel.js#L32-L39`](https://github.com/matrix-org/matrix-public-archive/blob/f95086df0d22393376916e087bc0ced09ba3f389/shared/viewmodels/DeveloperOptionsContentViewModel.js#L32-L39)

shared/views/RoomDirectoryView.js
Jem256: thank you for this suggestion. It has taught me about \b regex word-boundary character

shared/views/RoomDirectoryView.js
Jem256: No, the room list doesn't update. How would setting the boolean on the room object enable an update in the list? I don't understand

MadLittleMods: I don't think it would directly but we probably need to do something like that. The views listen to changes that are emitted.

We probably need to make a `class Room extends EventEmitter` [similar to what Hydrogen does](https://github.com/vector-im/hydrogen-web/blob/a565d2a7d377b97a9d02b850311aaaa0b9f88283/src/matrix/room/BaseRoom.js) so we can update the the `isNsfw` property and emit an update for the view to listen and update on.

Jem256: I create a class Room extends Eventmitter in my last [commit](https://github.com/matrix-org/matrix-public-archive/pull/155/commits/11ca940fd22c82fa019e2dd468f3de7148f782ec) but I still can't figure out how the changes will be emitted between the class and the view

Jem256: There's also this [issue from Hydrogen](https://github.com/vector-im/hydrogen-web/issues/727) that I think relates to our problem

MadLittleMods: @Jem256 I'm no Hydrogen expert to be able to explain or know exactly what to do here. I'd just have to hack on it myself in order to give any guidance. You might get some good notes by asking in [Hydrogen room on Matrix](https://matrix.to/#/!nFhKyoXMWWhcbXdKcs:matrix.org?via=matrix.org&via=element.io&via=t2l.io) or checking out the Hydrogen code base to find a similar pattern and how they solved it.

MadLittleMods:
Closing as this feature just landed via https://github.com/matrix-org/matrix-public-archive/pull/208

@Jem256 Thanks for making the effort to try this out! Sorry, I couldn't come with more Hydrogen knowledge to make this more possible for you to accomplish 🙇

Issue #59, opened 2022-09-01
Follow tombstone and predecessor history


MadLittleMods:
Follow tombstone and predecessor history.

In order to go backwards properly:

 1. Fetch the [`org.matrix.msc3946.room_predecessor`](https://github.com/matrix-org/matrix-spec-proposals/pull/3946) state event in the room (fallback to predecessor value from `m.room.create` state event)
 1. ~~Do a `/timestamp_to_event?dir=b&ts=<before the create event>` to see if any history was imported in the room and follow that first~~
     - No longer necessary because [MSC2716 wasn't used for Gitter](https://github.com/matrix-org/matrix-spec-proposals/pull/2716#issuecomment-1487441010)
 3. When we exhausted the pagination of the room, go to the `predecessor` room

In order to go forwards properly:

 1. Fetch the `m.room.tombstone` state event in the room
 1. Should we display events that happened after the `m.room.tombstone` event? Maybe not as the room creator wishes to move to another room.
 1. If we're on a date past the `m.room.tombstone` event, go to the `replacement_room`
 

---

Do we have to worry about verifying the links here? (make sure the tombstone and predecessor point at each other) 

It doesn't seem like it because the archive isn't doing any visual merging and we're just following the hints of the desired room you're viewing.
 
 ### Dev notes
 
Test to make sure that follow logic works:

 - [x] Paginating to the beginning of a room without a predecessor will work (just display the empty view)
     - There are already tests for this (see "Jumping forwards and backwards")
 - [ ] Paginating to the beginning of a room with a predecessor will redirect to the predecessor
 - [ ] ~~Paginating to the beginning of a room with a predecessor and history imported before the room was created will exhaust the imported history before following the predecessor~~
     - No longer necessary because [MSC2716 wasn't used for Gitter](https://github.com/matrix-org/matrix-spec-proposals/pull/2716#issuecomment-1487441010)
 - [ ] Paginating to the time after the tombstone, will following the tombstone to the `replacement_room`

Test to make sure we can still view all of the messages in the room:

  - [ ] Test to make sure you can view messages on the day that tombstone happened
  - [ ] Test to make sure you can view the first message in the room

PR #167, opened 2023-04-10
Follow tombstone and predecessor history

MadLittleMods:
Follow tombstone and predecessor history

Fix https://github.com/matrix-org/matrix-public-archive/issues/59

Other updates:

 - Update tests to use `/roomid/room1/date/2022/01/03` format instead of trying to retrofit the weird alias stuff on there. Which also makes the fancy to actual URL utilities much more simple.
 - Update to specify `archiveMessageLimit` in the test case because pages have different number of events depending on if we are against a boundary, hidden events, etc.

### Dev notes

 - [x] Follow tombstone and predecessor history when visiting room on a given day
 - [x] Follow tombstone and predecessor history for `/jump`
 - [x] Write tests
 - [ ] ~~Edge case: Someone can create a circular reference which will cause us to keep redirecting the user around in a loop. Do we care?~~
    - Maybe not as the `ERR_TOO_MANY_REDIRECTS` in the browser will handle it
 - [x] Does it work in a Gitter-like historical room where we created the room with some date X and imported history with timestamps before X?
    - Do we need some causality query parameter for `/timestamp_to_event`? Like `/timestamp_to_event?dir=f&ts=123&event_id=$123` meaning find the closest event from `ts` looking forward after `event_id` (the causality part, "A after B")
       - Perhaps we could just see some causality by observing `m.room.create` popping up in `/timestamp_to_event?dir=f` when we `/jump`. If we see `m.room.create`, then assume we've reached the end of the room and move to the successor
       - Someone could also tie us in knots by making timestamps circular so we probably do need a causality param for `/timestamp_to_event`
       - Created [MSC3999](https://github.com/matrix-org/matrix-spec-proposals/pull/3999) to track adding a causal query parameter to `/timestamp_to_event?event_id=$abc`


server/lib/matrix-utils/ensure-room-joined.js
```diff
@@ -4,15 +4,29 @@ const assert = require('assert');
...
+
+  // Let's do our best for the user to join the room. Since room ID's are
+  // unroutable on their own and won't be found if the server doesn't already
+  // know about the room, we'll try to join the room via the server name that
+  // we derived from the room ID or alias.
+  const viaServersWithAssumptions = new Set(viaServers);
+  const derivedServerName = getServerNameFromMatrixRoomIdOrAlias(roomIdOrAlias);
+  if (derivedServerName) {
+    viaServersWithAssumptions.add(derivedServerName);
+  }
```
MadLittleMods, line 26: TODO: Split this off into a separate PR

server/routes/room-routes.js
```diff
@@ -265,39 +282,84 @@ router.get(
...
-      // in the displayed range, say `T12:00:05` would still give us the same day
-      // `/2020/01/02` and we want to redirect them to previous chunk from that same
-      // day, like `/2020/01/02T12:00:00`
+      // in the displayed range (say that occured on `T12:00:25`) would still give us
+      // the same day `/2020/01/02` and we want to redirect them to previous chunk from
+      // that same day that still encompasses the closest message looking backwards,
+      // like `/2020/01/02T13:00:00`
       if (dir === DIRECTION.backward) {
+        // We choose `currentRangeStartTs` instead of `ts` (the jump point) because TODO: why?
+        // and we don't choose `currentRangeEndTs` because TODO: why?
```
MadLittleMods, line 352: I feel like I can't justify this 🤔 but it does seem correct from tested behavior.

I think my reasoning for using `currentRangeEndTs` before was that we could keep paginating with day precision if we were showing more than a day for each page. But then this falls apart when you have a bunch of  messages from the same hour that fill up multiple pages.

Using `ts` makes various tests fail with various precision differences. Probably because of the `- 1`/`+ 1` stuff that we want to remove in the future anyway when MSC3999 lands.

But those don't seem like explanations for why (the truth kernel)

MadLittleMods: I've left a link to this discussion thread so people can get some context. But I think I will opt to merging this PR without figuring this out. Will just have to contemplate this more and fill it out when an epiphany happens.

PR #175, opened 2023-04-19
Production ready build

MadLittleMods:
Production ready build

 - Rename `public` -> `client` so it doesn't get copied automagically as-is (without hashes which we want for cache busting), https://vitejs.dev/guide/assets.html#the-public-directory
     - We still build the version files to `public/` so their copied as-is and Vite handles it for us (so we can use `emptyOutDir`) 
 - Use a multiple entrypoint `.js` Vite build so things can be more intelligently bundled and take less time
     - We aren't using library mode because it doesn't minify or bundle assets
 - Using hash asset tags for cache busting. Hash of the file included in the file name
 - We lookup these hashed assets from `manifest.json` that Vite builds (https://vitejs.dev/guide/backend-integration.html) to serve and preload
 - In terms of optimized bundles, I know the current output isn't great now but will have to opt to fix that up separately in the future. Tracked by https://github.com/matrix-org/matrix-public-archive/issues/176 

 
### Dev notes

Cache busting:

 - https://www.alainschlesser.com/bust-cache-content-hash/

Gulp:

 - https://github.com/sindresorhus/gulp-rev
 - https://github.com/TheDancingCode/gulp-rev-rewrite

 Vite:
 
 - https://vitejs.dev/guide/backend-integration.html
 - https://vitejs.dev/guide/build.html#chunking-strategy
    - `manualChunks` and `experimentalMinChunkSize`
 - How can we minify the JS?
    - Can't minify with library mode: https://github.com/vitejs/vite/issues/5167
    - **Update:** We can stop using library (`lib`) mode in favor of a `.js` entry point (see final state of this PR)
 - How can we minify our CSS?
    - **Update:** We can stop using library (`lib`) mode in favor of a `.js` entry point (see final state of this PR)
 - How can we transform our assets (CSS, images, fonts) to put the file hash in the name?
    - Can't bundle assets in library mode: https://github.com/vitejs/vite/issues/3295
    - **Update:** We can stop using library (`lib`) mode in favor of a `.js` entry point (see final state of this PR)
    - List of plugins available: https://vitejs.dev/plugins/
       - https://github.com/vitejs/awesome-vite#plugins
       - https://vite-rollup-plugins.patak.dev/
    - Maybe [`rollup-plugin-emit-files`](https://github.com/juliendargelos/rollup-plugin-emit-files) could help
       - I see a custom [`emit-files-plugin.js`](https://github.com/GoogleChromeLabs/squoosh/blob/ecc715fe557ce57ba57f2535e87d68411bbf7de2/lib/emit-files-plugin.js) used in the squoosh project
    - [`vite-plugin-static-copy`](https://github.com/sapphi-red/vite-plugin-static-copy) and [`vite-plugin-copy-files`](https://github.com/mistjs/vite-plugin-copy-files) just copy files and don't transform/process anything

Other reference apps:

 - https://github.com/GoogleChromeLabs/squoosh


gulpfile.js
```diff
@@ -0,0 +1,160 @@
...
+    },
+  });
+}
+
+function processWatchServer() {
+  // Pass through some args
+  const args = [];
+  if (process.argv.includes('--tracing')) {
+    args.push('--tracing');
+  }
```
MadLittleMods, line 110: TODO: Add

```
if (process.argv.includes('--logOutputFromChildProcesses')) {
  args.push('--logOutputFromChildProcesses');
}
```

client/js/entry-client-hydrogen.js
```diff
@@ -0,0 +1,9 @@
+import 'matrix-public-archive-shared/hydrogen-vm-render-script';
```
MadLittleMods, line 1: The client scripts don't load in the browser. Trying to fix this up in https://github.com/matrix-org/matrix-public-archive/pull/175

PR #177, opened 2023-04-21
Refactor codebase from CommonJS to ESM

MadLittleMods:
Refactor codebase from CommonJS to ESM

Spawning from the `gulp-rev` dependency in https://github.com/matrix-org/matrix-public-archive/pull/175 being ESM which kinda necessitates that we also switch over.

### Dev notes

See https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

### DIFF
--- a/server/hydrogen-render/render-hydrogen-to-string-unsafe.js
+++ b/server/hydrogen-render/render-hydrogen-to-string-unsafe.js
@@ -12,7 +12,7 @@
 import assert from 'assert';
 import vm from 'vm';
 import path from 'path';
-const { readFile } = require('fs').promises;
+import { readFile } from 'node:fs/promises';
 import crypto from 'crypto';
 import { parseHTML } from 'linkedom';
 
--- a/server/hydrogen-render/render-hydrogen-to-string-unsafe.js
+++ b/server/hydrogen-render/render-hydrogen-to-string-unsafe.js
@@ -51,8 +51,6 @@ function createDomAndSetupVmContext() {
     vmContext.global.crypto = crypto.webcrypto;
   }
 
-  // So require(...) works in the vm
-  vmContext.global.require = require;
   // So we can see logs from the underlying vm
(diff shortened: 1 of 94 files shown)


server/hydrogen-render/render-hydrogen-to-string-unsafe.js
```diff
@@ -102,19 +98,43 @@ async function _renderHydrogenToStringUnsafe(renderOptions) {
...
+      );
+    });
   });
-  // Note: The VM does not exit after the result is returned here and is why
-  // this should be run in a `child_process` that we can exit.
-  const vmResult = hydrogenRenderScript.runInContext(vmContext);
-  // Wait for everything to render
-  // (waiting on the promise returned from the VM render script)
-  await vmResult;
+  await hydrogenRenderModule.evaluate();
```
MadLittleMods, line 126: This area will be the downfall of this migration 😢

With the previous `new vm.Script(code).runInContext(vmContext)`, we see `SyntaxError: Cannot use import statement outside a module`

Which prompted an update to `new vm.SourceTextModule(...) ...` but this [API surface](https://nodejs.org/docs/latest-v18.x/api/vm.html#class-vmmodule) is crazy and expects to implement our own `linker` when I just want default Node.js behavior.

 - https://github.com/nodejs/node/issues/31234
 - https://github.com/nodejs/node/issues/35848
 - https://github.com/jestjs/jest/blob/6e95bdcbf9e5c2e354f277d0db6e4dff1ad2c43f/packages/jest-runtime/src/index.ts#L474-L513
 
The `new vm.Script(...)` API should handle all of the complexity of whether I used `import` or `require` syntax and just load dependencies (sane defaults please).

MadLittleMods: (giving up for now ⏩)

PR #184, opened 2023-04-25
Make sure to restart the server after Vite `manifest.json` changes

MadLittleMods:
Make sure to restart the server after Vite `manifest.json` changes so it can pick up the latest and serve pages correctly.

### DIFF
--- a/server/start-dev.js
+++ b/server/start-dev.js
@@ -13,9 +13,7 @@ buildClient({
   build: {
     // Rebuild when we see changes
     // https://rollupjs.org/guide/en/#watch-options
-    watch: {
-      exclude: ['server/**/*'],
-    },
+    watch: true,
   },
 });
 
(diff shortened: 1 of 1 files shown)


server/start-dev.js
```diff
@@ -2,7 +2,7 @@
 
 console.log('start-dev process.env.NODE_ENV', process.env.NODE_ENV);
 
-const path = require('path');
+const path = require('path').posix;
```
MadLittleMods, line 5: Using the `posix` version to get the glob patterns to work. Output with the `verbose` option with `nodemon`:

Before:

```
Nodemon logs: detail: ignoring: .\.git/**/* .\dist\**\!(manifest.json)
```

After:

```
Nodemon logs: detail: ignoring: .\.git/**/* dist/**/!(manifest.json)
```

I'm curious if the `.git` pattern works here since it has the same backslash problem but 🤷, it just means extra retarts and I haven't noticed anything. How much git operations are you ever doing for this to matter anyway.

Issue #132, opened 2022-11-10
Page speed optimization - fast assets


MadLittleMods:
The page takes long enough load, we shouldn't need to additionally slow it down with 2 additional seconds of bootstrapping once it reaches the client.

![](https://user-images.githubusercontent.com/558581/200997265-9a048bd7-dec6-4c00-aa7a-d16ada6c42c9.png)

#### 1. Use a CDN to serve the assets fast

**Update:** The [production deployment](https://github.com/vector-im/sre-internal/issues/2079) will use Cloudflare (should be good)

It shouldn't take 2 seconds to serve 991kb JavaScript file (`entry-client-hydrogen.es.js`) for example.

We could speed up response times by using a CDN instead.

#### 2. Smaller assets

**Update:** Tracking separately as https://github.com/matrix-org/matrix-public-archive/issues/176

Almost 1 MB JavaScript file seems a little bigger than necessary.

One thing we can do is split Hydrogen out to its own bundle file so we can share it between the room directory and room archive view.

#### 3. `<Link rel="preload">` preload hints

**Update:** Added `Link` preload headers which are equivalent in https://github.com/matrix-org/matrix-public-archive/pull/171 https://github.com/matrix-org/matrix-public-archive/pull/185, and https://github.com/matrix-org/matrix-public-archive/pull/187

Use `<Link rel="preload" ...>` for the CSS, JavaScript, and fonts that we always have to load. This should also help the essential assets beat all of the image avatars that load before and block them as well currently.

Docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preload

```html
<head>
  <meta charset="utf-8" />
  <title>JS and CSS preload example</title>

  <link rel="preload" href="style.css" as="style" />
  <link rel="preload" href="main.js" as="script" />

  <link rel="stylesheet" href="style.css" />
</head>

<body>
  <h1>bouncing balls</h1>
  <canvas></canvas>

  <script src="main.js" defer></script>
</body>
```

#### 4. Better preload hints, HTTP 103

**Update:** See https://github.com/matrix-org/matrix-public-archive/issues/32 -> https://github.com/matrix-org/matrix-public-archive/pull/171

We can make the preload hints happen even sooner before the page even gets a HTML response. I think this could be a major win since it takes a while for us to assemble a response in the first place.

MadLittleMods:
[...]

PR #187, opened 2023-04-26
Preload fonts and images

MadLittleMods:
Preload fonts and images

Part of https://github.com/matrix-org/matrix-public-archive/issues/132

### TODO

 - [x] Make sure we don't duplicate load the fonts because the path comes to us as `Inter-Regular.woff2?v=3.13` with a query parameter which counts as a different URL than what we preloaded, see https://github.com/matrix-org/matrix-public-archive/pull/187#discussion_r1177638781

### DIFF
--- a/server/lib/set-headers-to-preload-assets.js
+++ b/server/lib/set-headers-to-preload-assets.js
@@ -16,13 +16,17 @@ function setHeadersToPreloadAssets(res, pageOptions) {
   assert(pageOptions);
   assert(pageOptions.entryPoint);
 
-  const { styles, preloadScripts } = getDependenciesForEntryPointName(pageOptions.entryPoint);
+  const { styles, fonts, images, preloadScripts } = getDependenciesForEntryPointName(
+    pageOptions.entryPoint
+  );
 
+  // Work on assembling the `Link` headers
+  //
   // Note: Any docs for the `<link>` element apply to the `Link` header. "The `Link`
   // header contains parameters [that] are equivalent to attributes of the `<link>`
   // element."
   // (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link#parameters)
-
+  //
   // XXX: Should we add `nopush` to the `Link` headers here? Many servers initiate an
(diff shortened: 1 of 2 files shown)


server/lib/set-headers-to-preload-assets.js
```diff
@@ -32,16 +36,20 @@ function setHeadersToPreloadAssets(res, pageOptions) {
...
   // We use `crossorigin` because fonts are fetched with anonymous mode "cors" and
   // "same-origin" credentials mode (see
   // https://drafts.csswg.org/css-fonts/#font-fetching-requirements). `crossorigin` is
   // just short-hand for `crossorigin=anonymous` (see
   // https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin and
   // https://html.spec.whatwg.org/multipage/infrastructure.html#cors-settings-attribute).
-  //
-  // `Link: <font_to_load.woff2>; rel=preload; as=font; crossorigin`
+  const fontLinks = fonts.map((fontUrl) => {
+    return `<${fontUrl}>; rel=preload; as=font; crossorigin`;
```
MadLittleMods, line 46: This works to preload but we currently preload a set of fonts that isn't used becayse the Hydrogen CSS has a query parameter added to font URL in the CSS which counts as a different URL than what we preloaded.

MadLittleMods: One potential solution is to use a PostCSS plugin to remove the query parameter from the CSS since we already include file hashes in the file name.

MadLittleMods: It looks like the Hydrogen crew already ran into this and removed the query parameter from their source via https://github.com/vector-im/hydrogen-web/pull/961

So we just need to update our [Hydrogen scratch branch](https://github.com/vector-im/hydrogen-web/pull/653) with the latest changes from Hydrogen and make a release ⏩

MadLittleMods: Tested and we only request one set of fonts now ✅

PR #194, opened 2023-04-26
Update docs in preparation for Matrix Public Archive being generally available 

MadLittleMods:
Update docs in preparation for Matrix Public Archive being generally available 

### Dev notes

App screenshot: `.../!OvgDmYaPOFRRWnIdQe:half-shot.uk/date/2021/08/06` at 90% zoom with even dimensions for better scaling 

Previous PR where we added the screenshot: https://github.com/matrix-org/matrix-public-archive/pull/42


docs/testing.md
```diff
@@ -4,12 +4,15 @@
 
 If you haven't setup `matrix-public-archive` yet, see the [_Setup_ section in the root `README.md`](../README.md#setup)
 
+Then we need to setup the federation cluster of homservers that we will test with.
```
pdurbin, line 7:
```suggestion
Then we need to setup the federation cluster of homeservers that we will test with.
```

docs/faq.md
```diff
@@ -0,0 +1,56 @@
+# FAQ
+
+## Can I run my own instance?
+
+Yes! We host a public canonical version of the Matrix Public Archive at
+[archive.matrix.org](https://archive.matrix.org/) that everyone can use but feel free to
+also run your own instance (setup instructions in the [readme](../README.md)).
+
+## How do I provide via servers to join from?
```
pdurbin, line 9: Are "via servers" a thing in Matrix? I'm confused by this question. 🤔

pdurbin (commented):
Looks great! I left a couple comments.

docs/faq.md
MadLittleMods: If your server doesn't know about a given room ID, then it won't be able to join it (room ID's aren't routable on their own). Which is where "via" servers come in and you provide a server that already knows about the room to join through. See the [`server_name` query parameter with the `/join` endpoint](https://spec.matrix.org/v1.5/client-server-api/#post_matrixclientv3joinroomidoralias).

Maybe this detail doesn't help much in reality. I can remove it for now until I can phrase it in a useful way.

MadLittleMods:
Thanks for the review @pdurbin 🐮

PR #195, opened 2023-04-27
Fix typos where I misspelled "homeserver"

MadLittleMods:
Fix typos where I misspelled `homserver` -> `homeserver`

Spawning from another typo that [@pdurbin pointed out in the docs](https://github.com/matrix-org/matrix-public-archive/pull/194#discussion_r1178510694).

### DIFF
--- a/shared/viewmodels/RoomDirectoryViewModel.js
+++ b/shared/viewmodels/RoomDirectoryViewModel.js
@@ -9,7 +9,7 @@ const HomeserverSelectionModalContentViewModel = require('matrix-public-archive-
 
 const DEFAULT_SERVER_LIST = ['matrix.org', 'gitter.im', 'libera.chat'];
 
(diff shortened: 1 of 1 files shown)


shared/viewmodels/RoomDirectoryViewModel.js
```diff
@@ -9,7 +9,7 @@ const HomeserverSelectionModalContentViewModel = require('matrix-public-archive-
 
 const DEFAULT_SERVER_LIST = ['matrix.org', 'gitter.im', 'libera.chat'];
 
-const ADDED_HOMESERVERS_LIST_LOCAL_STORAGE_KEY = 'addedHomservers';
+const ADDED_HOMESERVERS_LIST_LOCAL_STORAGE_KEY = 'addedHomeservers';
```
MadLittleMods, line 12: This will result in added homeservers essentially being reset since we're reading and storing under a new key but we haven't even cut a release yet so it's fine ⏩

PR #203, opened 2023-05-01
Fix styles on timeout page

MadLittleMods:
Fix styles on timeout page since we started using the [`manifest.json` for asset paths](https://github.com/matrix-org/matrix-public-archive/pull/175).

### DIFF
--- a/client/js/entry-client-timeout.js
+++ b/client/js/entry-client-timeout.js
@@ -1,0 +1,3 @@
+// Assets
+import 'hydrogen-view-sdk/assets/theme-element-light.css';
+import '../css/styles.css';
--- a/server/hydrogen-render/render-page-html.js
+++ b/server/hydrogen-render/render-page-html.js
@@ -8,7 +8,7 @@ const safeJson = require('../lib/safe-json');
 const getDependenciesForEntryPointName = require('../lib/get-dependencies-for-entry-point-name');
 const getFaviconAssetUrls = require('../lib/get-favicon-asset-urls');
 
-async function renderPageHtml({
+function renderPageHtml({
   pageOptions,
   // Make sure you sanitize this before passing it to us
   bodyHtml,
--- a/server/middleware/timeout-middleware.js
+++ b/server/middleware/timeout-middleware.js
@@ -5,7 +5,7 @@ const urlJoin = require('url-join');
 const asyncHandler = require('../lib/express-async-handler');
 const { getSerializableSpans, getActiveTraceId } = require('../tracing/tracing-middleware');
 const sanitizeHtml = require('../lib/sanitize-html');
-const safeJson = require('../lib/safe-json');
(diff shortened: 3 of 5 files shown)


server/middleware/timeout-middleware.js
```diff
@@ -5,7 +5,7 @@ const urlJoin = require('url-join');
 const asyncHandler = require('../lib/express-async-handler');
 const { getSerializableSpans, getActiveTraceId } = require('../tracing/tracing-middleware');
 const sanitizeHtml = require('../lib/sanitize-html');
-const safeJson = require('../lib/safe-json');
+const renderPageHtml = require('../hydrogen-render/render-page-html');
```
MadLittleMods, line 8: Using standard `renderPageHtml(...)` function to render the page

server/hydrogen-render/render-page-html.js
```diff
@@ -8,7 +8,7 @@ const safeJson = require('../lib/safe-json');
 const getDependenciesForEntryPointName = require('../lib/get-dependencies-for-entry-point-name');
 const getFaviconAssetUrls = require('../lib/get-favicon-asset-urls');
 
-async function renderPageHtml({
```
MadLittleMods, line 11: This doesn't need to be async and we actually used this incorrectly in `server/routes/client-side-room-alias-hash-redirect-route.js` since we never awaited the promise before.

No need to be async though so remove `async`

client/js/entry-client-timeout.js
```diff
@@ -0,0 +1,3 @@
+// Assets
+import 'hydrogen-view-sdk/assets/theme-element-light.css';
+import '../css/styles.css';
```
MadLittleMods, line 3: Adding an entrypoint just so we can pull in the necessary assets for the page and have it mapped in `manifest.json`

server/routes/client-side-room-alias-hash-redirect-route.js
```diff
@@ -24,16 +24,19 @@ function clientSideRoomAliasHashRedirectRoute(req, res) {
...
-      <span class="js-try-redirect-message" style="display: none">One sec while we try to redirect you to the right place.</span>
-    </h1>
-    <p>If there was a #room_alias:server hash in the URL, we tried redirecting  you to the right place.</p>
-    <p>
-      Otherwise, you're simply in a place that does not exist.
-      You can ${sanitizeHtml(`<a href="${basePath}">go back to the homepage</a>.`)}
-    </p>
-    `;
+    ${/* We add the .hydrogen class here just to get normal body styles */ ''}
+    <div class="hydrogen">
```
MadLittleMods, line 28: Restoring `hydrogen` here which was lost in https://github.com/matrix-org/matrix-public-archive/pull/175

Doesn't really matter for the redirect page though.

Issue #148, opened 2022-12-01
Apply timeout middleware to all routes


MadLittleMods:
Move the [`timeout-middleware`](https://github.com/matrix-org/matrix-public-archive/blob/2999691eea7d6dc97902d36d3e1cb59a0df923d6/server/routes/timeout-middleware.js) to apply to all routes.

It would also be good to send a metric whenever we see a request timeout, related to https://github.com/matrix-org/matrix-public-archive/issues/45

Adjacent: Here is an example middleware from the Gitter webapp that logs and metrics when a request is pending for more than 60 seconds, https://gitlab.com/gitterHQ/webapp/-/blob/676fadc3693260c8c51f448a0ca4c3e180d1b4a2/server/web/middlewares/pending-request.js#L50-84

Issue #40, opened 2022-07-13
Cancellable requests


MadLittleMods:
If someone cancels a long-running request in their browser, will we stop trying to talk to the Matrix homeserver? I assume we will continue to process the request as if nothing happened and seems to be the case according to https://stackoverflow.com/questions/35198208/handling-cancelled-request-with-express-node-js-and-angular

At the moment, the HTTP timeout for the app deployment in Red Hat OpenShift is 30 seconds while Synapse's timeout is 180 seconds (I think 120 seconds on `matrix.org`) so we often run into the scenario where the archive gets a 504 Gateway Timeout but are probably talking to the homeserver a lot more afterwards. Especially in some cases where every request is just slow (join, `/timestamp_to_event`, `/context`, `/messages`).

Synapse has recently been implementing cancellable requests so it will work on a granular level with them.

This is a minor thing for now so we can leave this for down the road.

## Dev notes

I've implemented a [cancellable request in Express once for the Gitter export resources](https://gitlab.com/gitterHQ/webapp/-/blob/615c78d0b0a314c2c9e4098b8d2ba0471d16961b/server/api-web/export/generate-export-resource.js#L84-87) since we stream out the data there and it can take a while.

```js
let isRequestCanceled = false;
req.on('close', function() {
  isRequestCanceled = true;
});
```

---

Notes:

 - `fetch` and `AbortSignal`, https://simonplend.com/how-to-cancel-an-http-request-in-node-js/

Issue #9, opened 2022-02-24
Request logging


MadLittleMods:
Basically we want per-request logging where logs that happen within a Express request can be correlated together. We need to store a request ID in the context of a request and be able to get the request ID without needing to pass around `request` everywhere.

 - https://www.izertis.com/en/-/blog-express-js-logging-info-with-global-unique-request-id-node-js
    - Talks about using [`express-http-context`](https://www.npmjs.com/package/express-http-context) and wrapping `winston` to include the request ID in the logs
    - Uses `cls-hooked` under the hood which is "Continuation-Local Storage" and the equivalent to "Thread Local Storage" (TLS) for Node.js in the call stack context (and `async_hooks`).
 - https://javascript.plainenglish.io/simple-request-specific-logging-for-node-js-212bb2c0e887
    - [`async_hooks`](https://nodejs.org/docs/latest-v16.x/api/async_hooks.html) are still "[Stability: 1](https://nodejs.org/docs/latest-v16.x/api/documentation.html#stability-index) - Experimental", 
    - Links to an interesting performance analysis about the performance hit when using `async_hooks`, https://github.com/bmeurer/async-hooks-performance-impact

PR #204, opened 2023-05-01
Timeout requests and stop processing further

MadLittleMods:
Timeout requests and stop processing further

Fix https://github.com/matrix-org/matrix-public-archive/issues/148
Fix https://github.com/matrix-org/matrix-public-archive/issues/40

 - Apply timeout middleware to all room directory and room routes
 - Stop messing with the response after we timeout. Fix https://github.com/matrix-org/matrix-public-archive/issues/148
    - This also involves cancelling any `async/await` things like requests in the routes so we throw an abort error instead of continuing on. Fix https://github.com/matrix-org/matrix-public-archive/issues/40
 - Also abort the route if we see that the user closed the request before we could respond to them
 - Bumps minimum supported Node.js version to v18 because we're now using the built-in native `fetch` in Node.js vs `node-fetch`. This gives us the custom `signal.reason` that we aborted with instead of a generic `AbortError`.
    - This also means we had to add some instrumentation for `fetch` which uses `undici` under the hood. Settled on some unofficial instrumentation: [`opentelemetry-instrumentation-fetch-node`](https://www.npmjs.com/package/opentelemetry-instrumentation-fetch-node)

### Dev notes

Avoid errors like the following where the route function still continued on after we sent a timeout response to the user:

```
Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at new NodeError (node:internal/errors:371:5)
    at ServerResponse.setHeader (node:_http_outgoing:576:11)
    at ServerResponse.header (matrix-public-archive/node_modules/express/lib/response.js:776:10)
    at ServerResponse.append (matrix-public-archive/node_modules/express/lib/response.js:737:15)
    at setHeadersToPreloadAssets (matrix-public-archive/server/lib/set-headers-to-preload-assets.js:73:7)
    at matrix-public-archive/server/routes/room-routes.js:908:5
```

Instead, we are left with these errors logged to the console each time we timeout. It would be best if we could ignore these in an error handler (added as a TODO here)

```
AbortError: The user aborted a request.
    at abort (matrix-public-archive/node_modules/node-fetch/lib/index.js:1448:16)
    at AbortSignal.abortAndFinalize (matrix-public-archive/node_modules/node-fetch/lib/index.js:1463:4)
    at AbortSignal.[nodejs.internal.kHybridDispatch] (node:internal/event_target:647:20)
    at AbortSignal.dispatchEvent (node:internal/event_target:589:26)
    at abortSignal (node:internal/abort_controller:283:10)
    at AbortController.abort (node:internal/abort_controller:314:5)
    at Timeout.<anonymous> (matrix-public-archive/server/middleware/timeout-middleware.js:25:25)
    at listOnTimeout (node:internal/timers:559:17)
    at processTimers (node:internal/timers:502:7)
```

Update: Switched to custom errors and native `fetch` in Node.js which emit the custom `signal.reason` instead of a generic error. So we now get a lovely error like this:

```
RouteTimeoutAbortError: Timed out after 25000ms while trying to respond to route /r/too-many-messages-on-day:my.synapse.server/date/2022/11/16
    at Timeout.<anonymous> (server/middleware/timeout-middleware.js:28:7)
    at listOnTimeout (node:internal/timers:569:17)
    at process.processTimers (node:internal/timers:512:7)
```

### Todo

 - [x] Support `abortSignal` in `renderHydrogenToString(...)`
 - [x] Add an error handler to ignore `AbortError` (see dev notes for example) which appear in the logs whenever we timeout. We don't need to see these in the logs


server/lib/fetch-endpoint.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-const fetch = require('node-fetch');
```
MadLittleMods, line 3: Because we switched from `node-fetch` to native `fetch` built in to Node.js, this changed the underlying function doing the request and `@opentelemetry/instrumentation-http` no longer covers us.

We need to make sure the `fetch` is instrumented and update this line:

https://github.com/matrix-org/matrix-public-archive/blob/f3318446f8b71e2e43329ac599ac33f548835169/server/tracing/tracing-middleware.js#L57

Also make sure we serialize it correctly on the timeout page

MadLittleMods: I think we need some instrumentation for `undici`, see https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1021

MadLittleMods: `opentelemetry-instrumentation-fetch-node` seems to work ✅

Example span from a fetch request using opentelemetry-instrumentation-fetch-node`

I tried `opentelemetry-instrumentation-undici` first but it doesn't seem to add any spans to the traces 🤷

server/middleware/timeout-middleware.js
```diff
@@ -16,16 +19,29 @@ assert(requestTimeoutMs);
...
+  req.abortController = new AbortController();
+  req.abortSignal = req.abortController.signal;
+
   const timeoutId = setTimeout(() => {
+    // Signal to downstream middlewares/routes that they should stop processing/fetching
+    // things since we timed out (downstream consumers need to respect `req.abortSignal`)
+    req.abortController.abort(
+      new RouteTimeoutAbortError(
+        `Timed out after ${requestTimeoutMs}ms while trying to respond to route ${req.originalUrl}`
+      )
```
MadLittleMods, line 31: Maybe, it would be interesting to attach some more info about the long-running requests that took up all of the time (get the spans).

Future PR stuff though.

Perhaps, it would just be better to have per-request correlated logging and see the other info that way, https://github.com/matrix-org/matrix-public-archive/issues/9

server/child-process-runner/run-in-child-process.js
```diff
@@ -151,18 +184,12 @@ async function runInChildProcess(modulePath, runArguments, { timeout }) {
...
-            childErrors,
-            childStdErr
-          );
-          reject(
-            new RethrownError(
-              `Timed out while running ${modulePath} so we aborted the child process after ${timeout}ms. Any child errors? (${childErrors.length})`,
-              childErrorSummary
-            )
-          );
+          reject(abortController.signal.reason || err);
```
MadLittleMods, line 192: Created https://github.com/nodejs/node/issues/47814 to track when we can stop doing this workaround

PR #207, opened 2023-05-02
We can better detect static assets to avoid tracing nowadays

MadLittleMods:
We can better detect static assets to avoid tracing nowadays because all assets are served from `/assets` since https://github.com/matrix-org/matrix-public-archive/pull/175

### DIFF
--- a/server/tracing/tracing.js
+++ b/server/tracing/tracing.js
@@ -84,11 +84,7 @@ function startTracing() {
           // https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1034#issuecomment-1158435392
           ignoreIncomingRequestHook: (req) => {
             // Ignore spans from static assets.
-            //
-            // FIXME: Ideally, all of the assets would all be served under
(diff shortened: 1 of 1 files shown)


server/tracing/tracing.js
```diff
@@ -84,11 +84,7 @@ function startTracing() {
           // https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1034#issuecomment-1158435392
           ignoreIncomingRequestHook: (req) => {
             // Ignore spans from static assets.
-            //
-            // FIXME: Ideally, all of the assets would all be served under
-            // `/static/` so we could ignore that way instead. In Hydrogen, this
-            // is tracked by https://github.com/vector-im/hydrogen-web/issues/757
-            const isStaticAsset = !!req.url.match(/\.(css|js|svg|woff2)(\?.*?)?$/);
```
MadLittleMods, line 91: This previous regex was also missing `.ico` and sourcemaps (`.*.map`)

PR #208, opened 2023-05-02
Add safe search filter for NSFW rooms

MadLittleMods:
Add safe search filter for NSFW rooms

Alternative community PR (unfinished): https://github.com/matrix-org/matrix-public-archive/pull/155

Fix https://github.com/matrix-org/matrix-public-archive/issues/89

### Todo

 - [x] Add custom styles to safe search toggle
 - [ ] ~~Should we add safe search to the query parameters: `?safe-search=false`~~
    - On one-hand, it would be nice for NSFW-tolerant people to be able to link to room directory as they see it. This would also solve the flash of blocked to unblocked when the page loads.
    - On the other hand, it allows people to just fling people to some nasty content although they could just link the room anyway.
    - In any case, it's simpler not to do this for now and we can iterate when people actually speak up with an actual need/use-case.
 - [x] Add test to ensure NSFW room is blocked on page-load
 - [x] Update Hydrogen with latest scratch changes

### DIFF
--- a/test/e2e-tests.js
+++ b/test/e2e-tests.js
@@ -2452,6 +2452,62 @@ describe('matrix-public-archive', () => {
         // Assert that the rooms we searched for on remote hs2 are visible
         assert.deepStrictEqual(roomsOnPageWithSearch.sort(), [roomXId, roomYId].sort());
       });
+
+      it('Safe search blocks nsfw rooms by default', async () => {
+        const client = await getTestClientForHs(testMatrixServerUrl1);
+        // This is just an extra room to fill out the room directory and make sure
+        // that it does not appear when searching.
+        await createTestRoom(client);
+
+        // Then create some NSFW rooms we will find with search
+        //
+        // We use a `timeToken` so that we can namespace these rooms away from other
+        // test runs against the same homeserver
+        const timeToken = Date.now();
+        const roomPlanetPrefix = `planet-${timeToken}`;
+        const roomUranusId = await createTestRoom(client, {
+          // NSFW in title
+          name: `${roomPlanetPrefix}-uranus-nsfw`,
+        });
+        const roomMarsId = await createTestRoom(client, {
+          name: `${roomPlanetPrefix}-mars`,
+          // NSFW in room topic/description
+          topic: 'Get your ass to mars (NSFW)',
+        });
+
+        // Browse the room directory searching the room directory for those NSFW rooms
+        // (narrowing down results).
+        archiveUrl = matrixPublicArchiveURLCreator.roomDirectoryUrl({
+          searchTerm: roomPlanetPrefix,
+        });
+        const { data: roomDirectoryWithSearchPageHtml } = await fetchEndpointAsText(archiveUrl);
+        const domWithSearch = parseHTML(roomDirectoryWithSearchPageHtml);
+
+        const roomsCardsOnPageWithSearch = [
+          ...domWithSearch.document.querySelectorAll(`[data-testid="room-card"]`),
+        ];
+
+        // Assert that the rooms we searched for are on the page
+        const roomsIdsOnPageWithSearch = roomsCardsOnPageWithSearch.map((roomCardEl) => {
+          return roomCardEl.getAttribute('data-room-id');
+        });
+        assert.deepStrictEqual(roomsIdsOnPageWithSearch.sort(), [roomUranusId, roomMarsId].sort());
+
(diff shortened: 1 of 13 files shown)


test/e2e-tests.js
```diff
@@ -2452,6 +2452,62 @@ describe('matrix-public-archive', () => {
...
+
+        const roomsCardsOnPageWithSearch = [
+          ...domWithSearch.document.querySelectorAll(`[data-testid="room-card"]`),
+        ];
+
+        // Assert that the rooms we searched for are on the page
+        const roomsIdsOnPageWithSearch = roomsCardsOnPageWithSearch.map((roomCardEl) => {
+          return roomCardEl.getAttribute('data-room-id');
+        });
+        assert.deepStrictEqual(roomsIdsOnPageWithSearch.sort(), [roomUranusId, roomMarsId].sort());
```
MadLittleMods, line 2494: Why is this test flakey? Example failure: https://github.com/matrix-org/matrix-public-archive/actions/runs/4868717759/jobs/8682443906#step:7:203

[Synapse doesn't cache results for searches on `publicRooms`](https://github.com/matrix-org/synapse/blob/0e8aa2a1b28dfce374294450a015d18884c89d36/synapse/handlers/room_list.py#L98-L109) so it shouldn't be a stale results problem with other tests cases interfering.

MadLittleMods: I can actually reproduce consistently locally after re-pulling the `latest` Synapse image (minimal test run reproduction: `npm run test -- --grep "Room directory"`)

Just did some bisecting of Synapse versions to figure out where the Synapse change was introduced:

 - `v1.81.0`: Test fails consistently ❌
 - `v1.80.0`: Test fails consistently ❌
 - `v1.80.0rc1`: Test fails consistently ❌
 - `v1.79.0`: Working ✅
 - `v1.68.0`: Working ✅

I don't know which Synapse commit/PR specifically yet

MadLittleMods: I've done a bisect on Synapse commits and found the culprit -> https://github.com/matrix-org/synapse/pull/15229. Not totally obvious what's going wrong/different.

Bisect ([relevant page of commits](https://github.com/matrix-org/synapse/commits/develop?after=0e8aa2a1b28dfce374294450a015d18884c89d36+174&branch=develop&qualified_name=refs%2Fheads%2Fdevelop)):

 - `4953cd71dfbb1925dc4a477efd2ed48c2dfd70d6` (2023-03-16): Test fails consistently ❌ (locally built `matrixdotorg/synapse:4953cd71`)
`a1c986939444db9bbf2343224091c88d18b3f598` (2023-03-13): Test fails consistently ❌ (locally built `matrixdotorg/synapse:a1c98693`)
`be4ea209e8bf92fb3660807c1fe8ad3d7d05621f` (2023-03-08): Test fails consistently ❌ (locally built `matrixdotorg/synapse:be4ea209`)
`88efc75bab2849b7b1cee52770dea3cf9925b2e8` (2023-03-08): Working ✅ (locally built `matrixdotorg/synapse:88efc75b`)
`8314646cd3563d9aaaf8028c9e58989b4ed980ba` (2023-03-07): Working ✅ (locally built `matrixdotorg/synapse:8314646c`)

MadLittleMods: Doesn't seem to be a `visibility` issue as the rooms still end up as `preset_config=public_chat` and `visibility=public`.

Seems like a race condition because it passes if I add a delay after the room creation. And I would think this is a read-after-write consistency issue since we're not even using workers. Created https://github.com/matrix-org/synapse/issues/15526 to track this ⏩

MadLittleMods: I've pinned the Synapse version to `v1.79.0` before the regression for now ⏩

If we need to test with the latest Synapse, we will need to introduce some wait/retry complexity like this:

```js
// Try to avoid flakey tests where the homeserver hasn't added the rooms
// to the room directory yet
await waitForRoomIdsInHomeserverRoomDirectory([roomUranusId, roomMarsId]);
```

MadLittleMods: `waitForResultsInHomeserverRoomDirectory(...)` was added in https://github.com/matrix-org/matrix-public-archive/pull/276

```js
await waitForResultsInHomeserverRoomDirectory({
  client,
  searchTerm: 'room-foo',
});
```

It's not possible to search directly for a `room_id` (at least in Synapse) so I've opted to searching for the room name.

Issue #211, opened 2023-05-04
Time selector is showing for a day with less than the limit of messages


MadLittleMods:
Time selector is showing for a day with less than the limit of messages.

It's a room with only 3 actual messages in it.

![](https://user-images.githubusercontent.com/558581/236303314-c4ddf579-3549-4297-8b71-1915557eadf7.png)

MadLittleMods:
This is just because we only care about whether the messages on the page are from the same day

https://github.com/matrix-org/matrix-public-archive/blob/9b067f8637e71992400a908eab71c1050c68be74/shared/viewmodels/ArchiveRoomViewModel.js#L144-L151

We could update to check whether `events.length > archiveMessageLimit` as well

PR #213, opened 2023-05-05
Fix time selector showing when less than the page limit of messages

MadLittleMods:
Fix time selector showing when less than the page limit of messages

Fix https://github.com/matrix-org/matrix-public-archive/issues/211

### DIFF
--- a/test/e2e-tests.js
+++ b/test/e2e-tests.js
@@ -697,17 +697,23 @@ describe('matrix-public-archive', () => {
         });
 
         it('does not show time selector when all events from the same day but not over the limit', async () => {
-          // Set this low so we don't have to deal with many messages in the tests
-          config.set('archiveMessageLimit', 5);
+          // Set this low so we don't have to deal with many messages in the tests But
+          // high enough to encompass all of the primoridial room creation events +
+          // whatever messages we send in the room for this test.
+          config.set('archiveMessageLimit', 15);
 
           const client = await getTestClientForHs(testMatrixServerUrl1);
+          // FIXME: This test is flawed and needs MSC3997 to timestamp massage the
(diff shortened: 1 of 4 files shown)


test/e2e-tests.js
```diff
@@ -697,17 +697,23 @@ describe('matrix-public-archive', () => {
...
+          // Set this low so we don't have to deal with many messages in the tests But
+          // high enough to encompass all of the primoridial room creation events +
+          // whatever messages we send in the room for this test.
+          config.set('archiveMessageLimit', 15);
 
           const client = await getTestClientForHs(testMatrixServerUrl1);
+          // FIXME: This test is flawed and needs MSC3997 to timestamp massage the
+          // `/createRoom` events otherwise the `areTimestampsFromSameUtcDay(...)` will
+          // always be false because the create room events are from today vs the
+          // timestamp massaged messages we send below.
```
MadLittleMods, line 709: This test still doesn't cover us because of this flaw.

If we really wanted to, we could get around this by not timestamp massaging the message in the room and visiting the `now` date 🤷

test/e2e-tests.js
```diff
@@ -697,17 +697,23 @@ describe('matrix-public-archive', () => {
         });
 
         it('does not show time selector when all events from the same day but not over the limit', async () => {
-          // Set this low so we don't have to deal with many messages in the tests
-          config.set('archiveMessageLimit', 5);
+          // Set this low so we don't have to deal with many messages in the tests But
+          // high enough to encompass all of the primoridial room creation events +
+          // whatever messages we send in the room for this test.
+          config.set('archiveMessageLimit', 15);
```
MadLittleMods, line 703: Fixing one of the flaws here at least

PR #228, opened 2023-05-11
Prevent Cloudflare from overriding our own 504 timeout page

MadLittleMods:
Prevent Cloudflare from overriding our own 504 timeout page

Explored in https://gitlab.matrix.org/matrix-public-archive/deployment/-/issues/2 (internal deployment issue)

> Cloudflare returns an Cloudflare-branded HTTP 502 or 504 error when your origin web server responds with a standard HTTP 502 bad gateway or 504 gateway timeout error:
>
> *-- https://developers.cloudflare.com/support/troubleshooting/cloudflare-errors/troubleshooting-cloudflare-5xx-errors/#502504-from-your-origin-web-server*

The only way to disable this functionality is to have an Enterprise Cloudflare plan and use the `Enable Origin Error Pages` option:

> **Enable Origin Error Pages**
>
> When Origin Error Page is set to “On”, Cloudflare will proxy the 502 and 504 error pages directly from the origin.
>
> Requires Enterprise or higher

So instead of dealing with that headache, we're just working around this by responding with a 500 error when we timeout. Should be good enough I think. The user won't know any difference but may affect what Search Engines think. Not sure search engines care about the distinction since the page is slow to respond anyway which they punish.

### DIFF
--- a/server/middleware/timeout-middleware.js
+++ b/server/middleware/timeout-middleware.js
@@ -102,8 +102,17 @@ async function timeoutMiddleware(req, res, next) {
       },
     });
 
-    // 504 Gateway timeout
-    res.status(504);
+    // The most semantic HTTP status code to return here is a 504 Gateway timeout but if
+    // you use Cloudflare in front of the archive, it will serve its own
+    // Cloudflare-branded 504 page if your own origin server responds with a 504. And
+    // the only way to disable this functionality is to have an Enterprise Cloudflare
+    // plan. So to workaround this, we return a 500 instead. Relevant Cloudflare docs:
+    // https://developers.cloudflare.com/support/troubleshooting/cloudflare-errors/troubleshooting-cloudflare-5xx-errors/#502504-from-your-origin-web-server
+    //
+    // We want to show our own timeout page because it has more information about what
+    // went wrong (e.g. which external Matrix API requests were slow).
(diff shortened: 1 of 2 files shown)


server/middleware/timeout-middleware.js
```diff
@@ -102,8 +102,17 @@ async function timeoutMiddleware(req, res, next) {
...
+    // The most semantic HTTP status code to return here is a 504 Gateway timeout but if
+    // you use Cloudflare in front of the archive, it will serve its own
+    // Cloudflare-branded 504 page if your own origin server responds with a 504. And
+    // the only way to disable this functionality is to have an Enterprise Cloudflare
+    // plan. So to workaround this, we return a 500 instead. Relevant Cloudflare docs:
+    // https://developers.cloudflare.com/support/troubleshooting/cloudflare-errors/troubleshooting-cloudflare-5xx-errors/#502504-from-your-origin-web-server
+    //
+    // We want to show our own timeout page because it has more information about what
+    // went wrong (e.g. which external Matrix API requests were slow).
+    res.status(config.workaroundCloudflare504TimeoutErrors ? 500 : 504);
```
MadLittleMods, line 114: This should be using `config.get('workaroundCloudflare504TimeoutErrors')`

Fixed in https://github.com/matrix-org/matrix-public-archive/pull/229

Issue #47, opened 2022-08-06
Opting Out


Sleuth56:
Has there been any thought given to how a room admins or homeserver admins could opt out their room or server?
My thought would be some sort of state event in a room and the already standard X-Robots-Tag for servers.

I know this project is in very early stages but given the distributed nature of the matrix network I believe this to be a very important thing, especially coming from the matrix core team (Or at least that's who it looks like it's being driven by).

MadLittleMods:
The archive will only access rooms where the history is `world_readable`. A room can be `public` without `world_readable` history so there is already mechanism to sorta control this.

But it is possible that we add an additional signal/control to determine whether search engines should index it. I imagine that you would still be able to access the room via Matrix public archive but we would tell Google not to index the room if that signal is present.

Mikaela:
Will removing world readability effect the archive or will the past history still be visible using it?

Is there any mechanism in Matrix for redacting the past public history that any room administrator can use without resorting to running bots/servers or unspecced hacks (`m.room.retention` and specced changing history visibility to members only and `/upgraderoom 10`)

Mikaela:
Another concern is do tombstoned rooms automatically get excluded by both this and matrix-static? They may be inaccessible if they have weird join rules and no one to invite.

MadLittleMods:
> Will removing world readability effect the archive or will the past history still be visible using it?

@Mikaela I'm not sure exactly how the `m.room.history_visibility` lookup will happen exactly (current state or state at the time). It depends on that.

> Is there any mechanism in Matrix for redacting the past public history [...]

Best to create an issue and ask elsewhere about this.

> Another concern is do tombstoned rooms automatically get excluded by both this and matrix-static? They may be inaccessible if they have weird join rules and no one to invite.

Any world readable rooms, including tombstoned. I don't have any details about the other project, [matrix-static](https://github.com/matrix-org/matrix-static).

MadLittleMods:
Related to https://github.com/matrix-org/synapse/issues/14127

Cyberes:
I have a few rooms that are set to `world_readable` and published to my homeserver's room directory that I don't want to be on my public archive site. My solution was to simply ban the archive bot from those rooms. This really isn't a good solution because the rooms are still listed and when you try to view them you get an error page.

What if the archive backend was to query all the rooms to check if its bot has access? If it is banned from a room then it would remove that room from the list on the website. It could also re-check when a user tries to view that room.

[...]

PR #239, opened 2023-05-30
Don't allow previewing `shared` history rooms

tulir:
Only `world_readable` can be considered as opting into having history publicly on the web. Anything else must not be ~~archived~~ viewable without login until there's a dedicated state event for opting into archiving.

See #47

### DIFF
--- a/server/lib/matrix-utils/fetch-room-data.js
+++ b/server/lib/matrix-utils/fetch-room-data.js
@@ -155,7 +155,6 @@ const fetchRoomData = traceFunction(async function (
     stateCanonicalAliasResDataOutcome,
     stateAvatarResDataOutcome,
     stateHistoryVisibilityResDataOutcome,
-    stateJoinRulesResDataOutcome,
     predecessorInfoOutcome,
     successorInfoOutcome,
   ] = await Promise.allSettled([
@@ -182,10 +181,6 @@ const fetchRoomData = traceFunction(async function (
         abortSignal,
       }
     ),
-    fetchEndpointAsJson(getStateEndpointForRoomIdAndEventType(roomId, 'm.room.join_rules'), {
-      accessToken: matrixAccessToken,
-      abortSignal,
-    }),
     fetchPredecessorInfo(matrixAccessToken, roomId, { abortSignal }),
     fetchSuccessorInfo(matrixAccessToken, roomId, { abortSignal }),
   ]);
@@ -220,12 +215,6 @@ const fetchRoomData = traceFunction(async function (
     historyVisibility = data?.content?.history_visibility;
   }
 
-  let joinRule;
-  if (stateJoinRulesResDataOutcome.reason === undefined) {
-    const { data } = stateJoinRulesResDataOutcome.value;
-    joinRule = data?.content?.join_rule;
-  }
-
   let roomCreationTs;
   let predecessorRoomId;
   let predecessorLastKnownEventId;
@@ -251,7 +240,6 @@ const fetchRoomData = traceFunction(async function (
     canonicalAlias,
     avatarUrl,
     historyVisibility,
-    joinRule,
     roomCreationTs,
     predecessorRoomId,
     predecessorLastKnownEventId,
--- a/server/routes/room-routes.js
+++ b/server/routes/room-routes.js
@@ -828,15 +828,13 @@ router.get(
       }),
     ]);
 
-    // Only `world_readable` or `shared` rooms that are `public` are viewable in the archive
-    const allowedToViewRoom =
-      roomData.historyVisibility === 'world_readable' ||
-      (roomData.historyVisibility === 'shared' && roomData.joinRule === 'public');
+    // Only `world_readable` rooms are viewable in the archive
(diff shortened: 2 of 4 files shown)


bkil:
@tulir Here's the thing. I want to refer to a certain verdict, certain discussion or a certain piece of user testimony that was made in a matrix room on another, external platform.

Do you feel that this should only be allowed via matrix.to that commands the recipient of my link to register an account (along with all the complexity) and install a client just to view such a message? People would not bother and share it on Reddit/Twitter/Facebook instead as they can easily link to that.

Or just more commonly screenshot the whole conversation! It is trendy to circulate such screenshots by just copy & pasting them disregarding all copyright and privacy concerns (in memes or as a replacement for cross-platform boosting). If your comment is included in such a screenshot (even just as context), you can't realistically do anything against it.

At the same time, if your comment was either pointed at or just included as context through such an ephemeral nojs-client permalink, you still have the option to rephrase or even redact the original post later on. If you ask me, this is both more privacy-friendly and more conductive towards helping decentralized FOSS alternatives spread.

server/routes/room-routes.js
```diff
@@ -828,15 +828,13 @@ router.get(
...
+    // Only `world_readable` rooms are viewable in the archive
+    const allowedToViewRoom = roomData.historyVisibility === 'world_readable';
 
     if (!allowedToViewRoom) {
       throw new StatusError(
         403,
-        `Only \`world_readable\` or \`shared\` rooms that are \`public\` can be viewed in the archive. ${roomData.id} has m.room.history_visiblity=${roomData.historyVisibility} m.room.join_rules=${roomData.joinRule}`
+        `Only \`world_readable\` rooms can be viewed in the archive. ${roomData.id} has m.room.history_visiblity=${roomData.historyVisibility} m.room.join_rules=${roomData.joinRule}`
       );
     }
```
MadLittleMods, line 840: @bkil Please create threads off the diff for discussion

server/routes/room-routes.js
```diff
@@ -828,15 +828,13 @@ router.get(
       }),
     ]);
 
-    // Only `world_readable` or `shared` rooms that are `public` are viewable in the archive
-    const allowedToViewRoom =
-      roomData.historyVisibility === 'world_readable' ||
-      (roomData.historyVisibility === 'shared' && roomData.joinRule === 'public');
```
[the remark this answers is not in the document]
MadLittleMods: We plan to move ahead with this PR to remove `shared` public rooms from the archive :fast_forward: 

The points in favor of keeping `shared` accessible can be summarized by the following but in the end, while the archive bot was respectful of the technical implications and doesn't expose messages to any further audience (random people), there is a social obligation to consider. This option has been represented as "members only" in many clients which doesn't leave any nuance.

Otherwise, the main idea was if I can view the messages from a Matrix client as a random user, I should also be able to see the messages in the archive. In both the native Matrix client and archive cases, it’s the same result when a random user wants to view a `shared` room:

 - A random Matrix accesses the room, they see the history
 - A random user accesses the room in the archive, they see the history
 - Search engines are not allowed in either case (that only applies to `world_readable` rooms)

The join is mostly a technical detail to anyone trying to view the room. While I don't think the join event provides much value to the room in the normal cases, it could have benefit in tracing bad actors for moderation.

From the [spec](https://spec.matrix.org/v1.7/client-server-api/#room-history-visibility):
> - `world_readable` - All events while this is the `m.room.history_visibility` value may be shared by any participating homeserver with anyone, regardless of whether they have ever joined the room.
> - `shared` - Previous events are always accessible to newly joined members. All events in the room are accessible, even those sent when the member was not a part of the room.

Removing `shared` rooms, does mean we’re re-introducing friction for a portion people which the archive eliminates (which homeserver do I choose, which client, why do I even need an account, how do I view this on mobile, how do I reference and share this message to someone not already in the Matrix ecosystem, etc). But people can update their room to be `world_readable` as they see fit now to regain these benefits for their community.

server/routes/room-routes.js
```diff
@@ -828,15 +828,13 @@ router.get(
       }),
```
MadLittleMods, line 828: @tulir I've made a few additional changes to the tests and surrounding code to make this PR more viable -> https://github.com/matrix-org/matrix-public-archive/pull/274 Sorry for moving ahead of review comments but it was hard to tell exactly what to do otherwise.

Feel free to [cherry-pick those commits](https://github.com/matrix-org/matrix-public-archive/pull/274) (or copy the changes here) so we can merge via this original PR with all of the context.

If you don't have any interest, I can also merge via that PR.

tulir: Copied the commits here

server/lib/matrix-utils/fetch-room-data.js
```diff
@@ -155,7 +155,6 @@ const fetchRoomData = traceFunction(async function (
     stateCanonicalAliasResDataOutcome,
```
MadLittleMods, line 155: Thanks for the contribution and patience @tulir :bow: :bird:

Issue #257, opened 2023-05-31
Set the display name of the bot to the web site for transparency


bkil:
Right now, this should be https://github.com/matrix-org/matrix-public-archive but after #219 is fixed, this could be simply https://archive.matrix.org

It might also be included in the user avatar as well if some platform fails to forward the display name.

The top of the web page and the readme should link to the FAQ that answers something like

> Why did `@archive:matrix.org` join my room?

## Background

Certain communities have a policy about only allowing bots that identify themselves, share their source code and the result of their activities. Some ask that bots respond to `!botlist`, `!rollcall` or if the bot is @-mentioned.

* https://tilde.chat/wiki/bots
* https://rfc.tildeverse.org/rfcs/2

## Alternatives

This could be achieved in multiple ways. The least intrusive of which is to make it trivial to find information about the bot, such as running it under its own subdomain & domain with information about it. The second best is what I suggested above. 

Another possibility is vocally announcing itself after joining, but some may actually find that creepier and more intrusive.

MadLittleMods:
Ideally, we'd have something like [extensible profiles](https://github.com/matrix-org/matrix-spec-proposals/issues/3795) to add a proper explanation and links but that isn't available yet.

The best we can probably do is update the display name. https://github.com/matrix-org/matrix-public-archive/issues/219 will be fixed in any case so we should be thinking of names in that context. Here are some names that have come up:

 - `archive.matrix.org observation assistant`
 - `Read-only user for archive.matrix.org`
 - `Non-archiving user for archive.matrix.org`
 - `Exploratory user for archive.matrix.org`
 - `archive.matrix.org permissions check`

It would be best if the name could preempt assumptions and wrap the kind of [explanation I gave yesterday](https://github.com/matrix-org/matrix-public-archive/issues/47#issuecomment-1568698809).

Having some avatar would be good but ideally would just be the [custom logo](https://github.com/matrix-org/matrix-public-archive/issues/94) of the Matrix Public Archive which doesn't exist yet.

---

I don't think we'd be able serve any bot commands since that would require it's own infrastructure just to answer a rollcall command explaining itself. Something that's really just better served by some profile information anyway.

Mikaela:
> archive.matrix.org observation assistant

I prefer this one as it clearly communicates that it's observing me and my room non-consentually.

jonaharagon:
[...]

PR #262, opened 2023-06-06
Add reason why the archive bot is joining the room

MadLittleMods:
Add reason why the archive bot is joining the room

Using the join `reason` added in [MSC2367](https://github.com/matrix-org/matrix-spec-proposals/pull/2367). Unfortunately, this PR doesn't have much effect because it doesn't look like many clients support it yet (Element doesn't support it for example).

Part of https://github.com/matrix-org/matrix-public-archive/issues/257

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -20,7 +20,7 @@ messages from any given date and day-by-day navigation.
 ## Why did the archive bot join my room?
 
 Only public Matrix rooms with `shared` or `world_readable` [history
-visibility](https://spec.matrix.org/v1.6/client-server-api/#room-history-visibility) are
+visibility](https://spec.matrix.org/v1.7/client-server-api/#room-history-visibility) are
 accessible in the Matrix Public Archive.
 
 But the archive bot (`@archive:matrix.org`) will join any public room because it doesn't
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -19,15 +19,15 @@ messages from any given date and day-by-day navigation.
 
 ## Why did the archive bot join my room?
 
-Only public Matrix rooms with `shared` or `world_readable` history visibility are
+Only public Matrix rooms with `shared` or `world_readable` [history
+visibility](https://spec.matrix.org/v1.6/client-server-api/#room-history-visibility) are
 accessible in the Matrix Public Archive.
 
 But the archive bot (`@archive:matrix.org`) will join any public room because it doesn't
 know the history visibility without first joining. Any room without `world_readable` or
-`shared` [history
-visibility](https://spec.matrix.org/v1.6/client-server-api/#room-history-visibility)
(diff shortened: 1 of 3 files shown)


docs/faq.md
```diff
@@ -19,15 +19,15 @@ messages from any given date and day-by-day navigation.
 
 ## Why did the archive bot join my room?
 
-Only public Matrix rooms with `shared` or `world_readable` history visibility are
+Only public Matrix rooms with `shared` or `world_readable` [history
+visibility](https://spec.matrix.org/v1.6/client-server-api/#room-history-visibility) are
```
Mikaela, line 23: Why specify an older version of spec instead of just latest or at least the current 1.7?

MTRNord:
> (Element doesn't support it for example).

I believe element android shows them. Not sure if its on join though.
Web for sure doesn't :/

server/lib/matrix-utils/ensure-room-joined.js
```diff
@@ -43,6 +43,14 @@ async function ensureRoomJoined(
       method: 'POST',
       accessToken,
       abortSignal,
+      body: {
+        reason:
+          `Joining room to check history visibility. ` +
+          `If your room is public with shared or world readable history visibility, ` +
```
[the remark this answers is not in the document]
MadLittleMods: I've updated the FAQ with the equivalents you might see in the UI but it's not possible to be exhaustive for how every client might expose these options. We're also trying to be brief and to the point with this join `reason` so I'm going to just let the pointer to the FAQ take the lead for people trying to understand more.

docs/faq.md
```diff
@@ -17,31 +17,55 @@ And with the introduction of the jump to date API via
...
-
-All public Matrix rooms are accessible to view in the Matrix Public Archive. But only
-rooms with history visibility set to `world_readable` are indexable by search engines.
-
-Also see https://github.com/matrix-org/matrix-public-archive/issues/47 to track better
-opt out controls.
+## Why did the archive bot join my room?
+
+Only public Matrix rooms with `shared` or `world_readable` [history
+visibility](https://spec.matrix.org/v1.7/client-server-api/#room-history-visibility) are
```
Mikaela, line 23: Is the spec expected to greatly change or break the link or will this be updated every time a new spec releases?

```suggestion
visibility](https://spec.matrix.org/latest/client-server-api/#room-history-visibility) are
```

MadLittleMods: It doesn't really matter in this case because `v1.6` and `v1.7` have the information we're trying to explain.

Using `latest` means we get any further clarifications added but also means that the link is probably going to break at some point.

I'm fine with using `latest` for now though :fast_forward:

MadLittleMods:
Thanks for the review @Mikaela and @MTRNord 🦘

Issue #271, opened 2023-06-22
Search should return private rooms where the bot is already in or invited to (`world_readable`)


bkil:
I made a private room where I invited the bot (or even have the bot joined in from the past) I expect that room search on the archive site should return this room as it is able to show its chat log.

----
_See below for some background information._

_If I invite the bot to a room that is invite-only but has history visibility set to public, it would be desirable to show the chat log via the web._

_You may ask why this odd combination is beneficial. It's a new proposal against scammers for public groups. Anything else other than invite-only allows for spamming and flooding room state, and we proposed creating a CAPTCHA bot that would let you speak this way, while still retaining readability of the room for outsiders._

https://archive.matrix.org/r/mod-ideas:matrix.org/date/2023/06/22?at=$Y42s_vlxoOy_JRm71sKx09a8gguAgK-fKjRe29FtQXk

MadLittleMods:
I think the current logic would work how you want if you set your history visibility to `world_readable`. Is this sufficient for your use case?

https://github.com/matrix-org/matrix-public-archive/blob/e4800852fff23cd4a03ef05282c71f58516cf2a4/server/routes/room-routes.js#L831-L834

Anything else seems a bit precarious to support.

Maybe, if we support a [`X-Robots-Tag`/`<meta name="robots" content="noindex, nofollow">`](https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag) like thing that is being proposed in [MSC4021](https://github.com/matrix-org/matrix-spec-proposals/pull/4021), we could also include an explicit `index` directive which is mentioned in a few places but is actually not a specified directive and assumed to be a default anyway ([reference](https://yoast.com/robots-meta-tags/))

bkil:
Muting people based on power level won't work. The moment you allow arbitrary joining in your room, it is vulnerable to uncontrolled join-flooding that poisons aggregate room state that is still not dealt with properly and making joins and sync slower for everyone.

Hence why I suggest to don't let them join in the first place.

I set history visibility to wide-open, the room to private and invited the bot, but the bot did not join the room.

MadLittleMods:
> I set history visibility to wide-open, the room to private and invited the bot, but the bot did not join the room.

@bkil I think if you try to visit the room in the archive, the bot user will try to join and may be successful now that you invited it.

bkil:
[...]

PR #272, opened 2023-06-23
Optimistically use Matrix API's and show world readable rooms

MadLittleMods:
Previously, we would try to join the room no matter what. Now with this PR, we assume that we are already joined to the room or can peek in the `world_readable` case and only if we get a `403 Forbidden`, do we go back and try to join the room.

This has the effect of Matrix Public Archive being able to serve content from `world_readable` rooms without joining. And in the case of a `world_readable` room not being public, will still work (see case mentioned in https://github.com/matrix-org/matrix-public-archive/issues/271)

We also show `world_readable` rooms in the room directory instead of filtering them out.

Fix https://github.com/matrix-org/matrix-public-archive/issues/271

### DIFF
--- a/server/lib/matrix-utils/create-retry-fn-if-not-joined.js
+++ b/server/lib/matrix-utils/create-retry-fn-if-not-joined.js
@@ -1,0 +1,94 @@
+'use strict';
+
+const assert = require('assert');
+
+const { HTTPResponseError } = require('../fetch-endpoint');
+const ensureRoomJoined = require('./ensure-room-joined');
+
+const JOIN_STATES = {
+  unknown: 'unknown',
+  joining: 'joining',
+  joined: 'joined',
+  failed: 'failed',
+};
+const joinStateValues = Object.values(JOIN_STATES);
+
+// Optimistically use the Matrix API assuming you're already joined to the room or
+// accessing a `world_readable` room that doesn't require joining. If we see a 403
+// Forbidden, then try joining the room and retrying the API call.
+//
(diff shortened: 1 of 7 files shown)


server/lib/matrix-utils/create-retry-fn-if-not-joined.js
```diff
@@ -0,0 +1,92 @@
...
+  assert(viaServers instanceof Set);
+
+  let joinState = JOIN_STATES.unknown;
+  let joinPromise = null;
+
+  return async function retryFnIfNotJoined(fn) {
+    assert(
+      joinStateValues.includes(joinState),
+      `Unexpected internal join state when using createRetryFnIfNotJoined(...) (joinState=${joinState}). ` +
+        `This is a bug in the Matrix Public Archive. Please report`
```
wojtekLs, line 39: Its unclear where one should report state. Please be concise. Add link perhaps....

server/lib/matrix-utils/create-retry-fn-if-not-joined.js
```diff
@@ -0,0 +1,92 @@
...
+    );
+
+    if (joinState === JOIN_STATES.joining) {
+      // Wait for the join to finish before trying
+      await joinPromise;
```
wojtekLs, line 44: await `null`? In line 33 you define `joinPromise` with the `null` value.

wojtekLs (changes requested):
Left some commentary.

PR #287, opened 2023-11-09
Add CODEOWNERS

davidegirardi:


wojtekLs (changes requested):
there seem to be an error on line 1.

davidegirardi:
It's a matter of permissions for the group.

